PR #27, opened 2017-04-03
Remove flow annotation

cmcewen:
Having a `// @flow` comment in the file without annotations will throw flow errors


DevHossamHassan:
@cmcewen Thank you :) 👍

PR #32, opened 2017-05-07
Fix setUserName typo

DevHossamHassan:
fix for issue https://github.com/Instabug/instabug-reactnative/issues/29


SudoPlz:
I think that that's still an issue.

On the 3.0 android instabug sdk it's setUsername.

Now you're calling setUserName and there's no such method.

Do we need to reverse the pull request to get it working now?

Thanks!

PR #49, opened 2017-07-18
:bug: Fix the bug where the sdk gets invoked automatically at app lau…

salmatarzi:
…nch, and need to press twice on a button to manually invoke the sdk, fix a bug where changing the value of the strings did not work correctly

### DIFF
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
@@ -2,6 +2,8 @@ package com.instabug.reactlibrary;
 
 import android.app.Application;
 import android.net.Uri;
+import android.os.Handler;
+import android.os.Looper;
 
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -722,6 +722,16 @@ module.exports = {
     },
 
     /**
+     * Enable/Disable debug logs from Instabug SDK
+     * Default state: disabled
+     *
+     * @param isDebugEnabled whether debug logs should be printed or not into LogCat
+     */
+    setDebugEnabled: function(isDebugEnabled) {
+        Instabug.setDebugEnabled(isDebugEnabled);
```
DevHossamHassan: @salmatarzi  This line will throw exception  in case we don't have the same API for iOS 
Please add it for iOS or add an 'if' condition to make sure that we are on the proper OS

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
```diff
@@ -1105,7 +1109,7 @@ public void run() {
 
     private InstabugCustomTextPlaceHolder.Key getStringToKeyConstant(String key) {
         String keyInLowerCase = key.toLowerCase();
-        switch (keyInLowerCase) {
+        switch (key) {
```
DevHossamHassan: Please revert this change  back to avoid typos

DevHossamHassan: Good catch @salmatarzi  👏 
Just Please remove the redundant line "      String keyInLowerCase = key.toLowerCase();"

PR #50, opened 2017-07-20
Salmat/test release

salmatarzi:


package.json
```diff
@@ -23,7 +23,7 @@
   "homepage": "https://github.com/Instabug/instabug-reactnative#readme",
   "rnpm": {
     "android": {
-      "packageInstance": "new RNInstabugReactnativePackage(\"YOUR_ANDROID_APPLICATION_TOKEN\",MainApplication.this,\"shake\",\"#1D82DC\")"
+      "packageInstance": "new RNInstabugReactnativePackage.Builder(\"YOUR_ANDROID_APPLICATION_TOKEN\",MainApplication.this)\n.setInvocationEvent(\"shake\")\n.build()"
```
DevHossamHassan: @salmatarzi I'd prefer to add all other options here with their default values.

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativePackage.java
```diff
@@ -26,7 +27,8 @@
     private InstabugColorTheme instabugColorTheme = InstabugColorTheme.InstabugColorThemeLight;
 
     public RNInstabugReactnativePackage(String androidApplicationToken, Application androidApplication,
-                                        String invocationEventValue, String primaryColor) {
+                                        String invocationEventValue, String primaryColor,
+                                        InstabugFloatingButtonEdge floatingButtonEdge, int offset) {
```
DevHossamHassan: @salmatarzi This will break users' apps, we should deprecate the old constructor and change its implementation to make use of this constructor and not to duplicate the code.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "instabug-reactnative",
-  "version": "1.1.6",
+  "version": "1.1.8",
```
DevHossamHassan: Please keep the version number as it was, Since we are going to pump it later on.

PR #58, opened 2017-08-17
:bug: Fix a bug where the setUserAttribute API in iOS mixed the key a…

salmatarzi:
…ttribute with the value and vice versa.


salmatarzi:
@yousefhamza Can you please review this PR?

PR #59, opened 2017-08-22
0.47 backward compatibile Android support

fcmatteo:
React Native 0.47 has removed the `createJSModules` method. Had to remove the `@Override` to make the lib compatible with that version of RN.

https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8


SudoPlz:
@yousefhamza can you also review that?

We need it in order to be able to use the latest version of react-native.

Thanks.

salmatarzi:
@fcmatteo @SudoPlz Thank you!

PR #62, opened 2017-08-30
Removed the postinstall script from package.json

SudoPlz:
I believe it's not a good idea to mix pods with npm.
That `postinstall` script can only cause trouble (i.e our CI build now fails because of this).
It's very easy to `pod install` how come did you decide to force it?


SudoPlz:
Apart from the CI failing to build, another problem I see with that is you force us to `pod install` on `postinstall` and so we don't get to decide when to `pod install` ourselves.

That means that if we decide to `pod install` later we'll have ran `pod install` more than once = huge build times for no reason.

SudoPlz:
https://github.com/Instabug/instabug-reactnative/pull/64

PR #65, opened 2017-08-31
Fix/add binary in library

salmatarzi:


salmatarzi:
@yousefhamza Can you review this PR?

SudoPlz:
I'm curious though, why did you choose to add the binary in the project?

salmatarzi:
@SudoPlz For mainly two reasons: 
1. To maintain the consistency between the React Native SDK, and the Native SDK
2. To be less dependent on Cocoapods in case someone isn't using it

PR #72, opened 2017-09-07
Rename iOS exported method 'logout' to 'logOut' to match Android implementation

nmartin867:
The "logOut" method should be exposed with the same name in both iOS and Android. Currently it is necessary to write platform specific JS code to handle both cases. This PR renames the iOS exposed 'logout' method to 'logOut' to match Android.

This would be a breaking change for future iOS versions. The method could be overloaded to provide backward compatibility. I'd be happy to update this PR in that case.


salmatarzi:
> This would be a breaking change for future iOS versions. The method could be overloaded to
> provide backward compatibility. I'd be happy to update this PR in that case.

In previous iOS versions, before this change, this method ```Instabug.logOut()``` would not be working, however now it shall work correctly for both platforms. There's no need to provide backward compatibility.

Issue #80, opened 2017-10-04
Calling `Instabug.setStringToKey` on fails on android


alliannas:
When setting up InstantBug with with custom strings for keys the app fails on android
![android_instant_bug](https://user-images.githubusercontent.com/1064453/31185464-8de0a48a-a91b-11e7-9ea4-30feca0bf02f.png)

Happens on

instanbug-rectnative version: 2.0.5 
react-native version: 0.48.3

Example call
```
Instabug.startWithToken(
  INSTABUG_TOKEN,
  Instabug.invocationEvent.shake,
  Instabug.setEmailFieldRequired(false),
  Instabug.setShakingThresholdForAndroid(450),
  Instabug.setPrimaryColor(processColor('00FF00')),
  Instabug.setStringToKey(
    'Enter your email (optional)',
    Instabug.strings.emailFieldHint
  )
);
```

reemalashry1:
@alliannas Thank you for the fix! 😊

alliannas:
No problem :) Thank for having the bindings open-source and making it possible to due fixes like this 👏

PR #81, opened 2017-10-04
Fixes setString method name for android

alliannas:
Fixes for the issue #80


salmatarzi:
@alliannas Thank you for noticing this! 🙏🏽

PR #83, opened 2017-10-09
Fix bad access to null variadic args

### LINKED ISSUE #82: EXC_BAD_ACCESS in InstabugReactLogFunction
I came across the same issue as reported in https://github.com/Instabug/instabug-reactnative/issues/77#issuecomment-335014604.

The issue is that since `InstabugReactLogFunction` is not a variadic function, `va_list arg_list` is null here: https://github.com/Instabug/instabug-reactnative/blob/7cc7e90/ios/RNInstabug/InstabugReactBridge.m

It looks like the implementation of `IBGNSLogWithLevel` expects this to be non-null, but I can't check that since the iOS SDK is closed source 😢 

I would recommend not using `va_list` here, since it will always be null, and thus provides no value.

eliperkins:
This removes the need to access `va_list` in a non-variadic function.

By adding a variadic logging function, we can still bridge `RCTLogFunction`, without changing the public API of `IBGNSLogWithLevel` to not require a `va_list`.

This also removes the need to call the log function on the main queue, which was applied in #76 (7cc7e90a5a49629519706942334c9e6e9957e628).

Fixes #82

### DIFF
--- a/ios/RNInstabug/InstabugReactBridge.m
+++ b/ios/RNInstabug/InstabugReactBridge.m
@@ -442,6 +442,15 @@ RCT_EXPORT_METHOD(isRunningLive:(RCTResponseSenderBlock)callback) {
     return [iOSVersion compare:[UIDevice currentDevice].systemVersion options:NSNumericSearch] == NSOrderedDescending;
 };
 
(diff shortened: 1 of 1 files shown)


salmatarzi:
Thank you @eliperkins for this! 🙏🏽

Issue #86, opened 2017-10-18
Build error: ld: framework not found Instabug


jozan:
I got this error while trying to build with custom schemes. Turned out that `react-native link instabug-reactnative` adds `../node_modules/instabug-reactnative/ios` only to the default schemes `Debug` and `Release`. Could it detect that there are custom schemes defined and handle those properly as well?

It's this line which does the trick: https://github.com/Instabug/instabug-reactnative/blob/master/link.rb#L40

PR #87, opened 2017-10-18
Add framework path to all build configurations

jozan:
Fix to #86 

Dunno if this should be done to `unlink.rb` as well. Ping me if that should be done as well.


salmatarzi:
@jozan Thank you for this 🙏🏽

jozan:
Np, thanks for merging!

PR #103, opened 2017-12-20
✨ Add setVideoRecordingFloatingButtonPosition for Android

salmatarzi:


android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
```diff
@@ -1211,6 +1232,26 @@ public void setEnableInAppNotificationSound(boolean shouldPlaySound) {
...
+            } else if (cornerValue.equals(TOP_LEFT)) {
+                corner = InstabugVideoRecordingButtonCorner.TOP_LEFT;
+            } else if (cornerValue.equals(TOP_RIGHT)) {
+                corner = InstabugVideoRecordingButtonCorner.TOP_RIGHT;
+            }
+
+            return corner;
+        } catch (Exception e) {
+            e.printStackTrace();
+            return corner;
```
DevHossamHassan: @salmatarzi it would be more elegant if you return the corner only once after the catch block (at the end of the method)

PR #111, opened 2018-01-06
Feature/new release apis

salmatarzi:


index.js
```diff
@@ -376,13 +399,10 @@ module.exports = {
      * @param {boolean} galleryImage A boolean to enable or disable gallery image
      * attachments. In iOS 10+,NSPhotoLibraryUsageDescription should be set in
      * info.plist to enable gallery image attachments.
-     * @param {boolean} voiceNote A boolean to enable or disable voice note attachments.
-     * In iOS 10+, NSMicrophoneUsageDescription should be set in info.plist to enable
-     * voiceNote attachments.
      * @param {boolean} screenRecording A boolean to enable or disable screen recording attachments.
      */
-    setAttachmentTypesEnabled: function (screenshot, extraScreenshot, galleryImage, voiceNote, screenRecording) {
-        Instabug.setAttachmentTypesEnabled(screenshot, extraScreenshot, galleryImage, voiceNote, screenRecording);
```
DevHossamHassan: @salmatarzi Salma This change will break users code because they are currently using it with 5 params 
Instead: You may need to create a new API and keep the old one and mark it as deprecated,
Don't forget to modify the old API to navigate to the new one.

index.js
```diff
@@ -382,7 +406,21 @@ module.exports = {
      * @param {boolean} screenRecording A boolean to enable or disable screen recording attachments.
      */
     setAttachmentTypesEnabled: function (screenshot, extraScreenshot, galleryImage, voiceNote, screenRecording) {
-        Instabug.setAttachmentTypesEnabled(screenshot, extraScreenshot, galleryImage, voiceNote, screenRecording);
+        Instabug.setEnabledAttachmentTypes(screenshot, extraScreenshot, galleryImage, screenRecording);
```
yousefhamza: Double check they r the same order in both of the native SDKs

yousefhamza: No need 👍

PR #171, opened 2018-06-25
adding header search paths - expo detached

stefangomez:
had to add these header search paths this to get instabug to build properly for a detached expo project. Problem is outlined here: https://docs.expo.io/versions/v28.0.0/expokit/expokit#ios

Tested this with a fresh expo detached build, expo v28 and the build now works. With current package and instructions, this would happen otherwise:
Build Error

Hopefully this helps others, or you can implement a better method for expo detached projects


HeshamMegid:
@stefangomez closing this PR for now since it shouldn't be needed after adding CocoaPods support. Please feel free to reopen if you think we still need it.

PR #210, opened 2018-10-16
Fix isRunningLive always returning true on device

tomauty:
Recommend merging this quickly. The isRunningLive feature is effectively broken for the RN -> iOS integration. It will always return true.


HeshamMegid:
Thanks for the fix! Will be releasing a new version shortly.

tomauty:
Sure :) Sorry about the git mess...realized my original commit said the opposite, tried amending, got that. Haha.

PR #217, opened 2018-11-09
Android crashes when calling setInvocationEvent() from Javascript

cpetzel:
When calling `setInvocationEvent()` in javascript, this message gets sent across the bridge on a non-ui thread. The native instabug internal Android code eventually touches the UI view hierarchy, which cannot be touched by a non-UI thread... 

This PR simply executes this internal method on the UI thread.


salmamali:
@cpetzel Thank you for the fix.

PR #237, opened 2019-01-18
Allow RN package initialization to configure crash reporting

cpetzel:
By default, the Instabug React Native Android SDK will enable crash reporting, which will set the default thread exception handler... If I want to use a different crash handler, I have no option of disabling this, until my javascript code executes `Instabug.setCrashReportingEnabled(false);
`. This means that there are potentially a few seconds of critical startup time where crashes will be sent to instabug. 

This PR lets the developer opt out of crash reporting when initializing the Instabug React Native package.


HeshamMegid:
@cpetzel thanks for the pull request. 

This makes sense. We will give it a round of testing then merge it.

HeshamMegid:
@cpetzel this is included in v8.0.23, which will be available shortly.

Issue #239, opened 2019-02-01
Instabug_dsym_upload.sh - os=ios missing in API call


sregg:
### Steps to Reproduce the Problem

Use `Instabug_dsym_upload.sh` in Xcode 10 Post Build Action

### Expected Behavior

dSym files are uploaded

### Actual Behavior

dSym files are not uploaded

### Instabug integration code
```
#- -- SCRIPT BEGIN - --
# SKIP_SIMULATOR_BUILDS=1
  SCRIPT_SRC=$(find "$PROJECT_DIR" -name 'Instabug_dsym_upload.sh' | head -1)
  if [ ! "${SCRIPT_SRC}" ]; then
    echo "Instabug: err: script not found. Make sure that you're including Instabug.bundle in your project directory"
    exit 1
  fi
  APP_TOKEN="MY-APP-TOKEN"
  source "${SCRIPT_SRC}"
#- -- SCRIPT END - --
```

### SDK Version
8.0.23

### React Native, iOS and Android Versions
R-N: 0.57.7
iOS: Xcode 10

### Comments

The issue is in `Instabug_dsym_upload.sh`, the `-F os=ios` is missing in the curl command to upload the zip.
Without it, the curl returns 500.
After adding it manually in `Instabug_dsym_upload.sh`, the dsym are successfully uploaded.

sregg:
I pushed a PR for that fix: https://github.com/Instabug/Instabug-React-Native/pull/240

PR #240, opened 2019-02-01
Fixed curl command in Instabug_dsym_upload.sh

sregg:
Added missing os=ios param

Fixes https://github.com/Instabug/Instabug-React-Native/issues/239


HeshamMegid:
@sregg thanks for the PR!

PR #252, opened 2019-03-07
fix some type define issue

pengcao1:
1 `setstringToKey` params `key` type should be `strings` enum NOT a `string`
2 `setPrimaryColor(setPrimaryColor: string)` should be process by `processColor` because IOS UIColor does not support color string


pengcao1:
`setstringToKey` should be `setStringToKey`

pengcao1:
please help to merge this PR ASAP.
a fatal issue is: 
Declaration the function signature in *.d.ts as `setstringToKey`
BUT, the truth of this function should be `setStringToKey`

PR #257, opened 2019-03-19
Don't mix import and module.exports

Empact:
Webpack disallows accessing `module.exports` from within es6 files, e.g.
where `import` is used. This assignment results in:
"Attempted to assign to readonly property"
https://github.com/webpack/webpack/issues/3491

### DIFF
--- a/index.js
+++ b/index.js
@@ -957,4 +957,4 @@ InstabugModule.Chats = Chats;
 InstabugModule.Replies = Replies;
 InstabugModule.CrashReporting = CrashReporting;
 
-module.exports = InstabugModule;
+export default InstabugModule;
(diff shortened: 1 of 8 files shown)


Empact:
We're using Webpack via Haul, incidentally: https://github.com/callstack/haul

HeshamMegid:
@Empact thanks for the pull request! We'll be merging this and including it in our next release.

soumyamishra89:
This merge has disallowed destructuring on import, so now 
import {BugReporting} from 'instabug-reactnative' does not work.
Instead one has to import Instabug from 'instabug-reactnative'
and do:  const {BugReporting} = Instabug to let everything work with previous version in case one had used destructuring on import.

Was this desired?, if yes then the docs needs to be updated. If no, then please add individual export to each class of Instabug.

Empact:
@soumyamishra89 not desired on my part. I'll make a follow-up today.

salmamali:
@Empact @soumyamishra89 This was not intended. We will release the fix shortly.

PR #277, opened 2019-05-02
Correction for Instabug dependency

edcrampin:
Instabug dependency in build.gradle listed as 8.2.2.0. This doesn't exist in Maven so update to 8.2.2


HeshamMegid:
@edcrampin we've switched away from Maven to our own repo available at https://sdks.instabug.com/nexus/repository/instabug-cp

Unlinking and linking again should add this automatically to your `build.gradle` file. You can also add it manually. Let me know if you have any issues with this.

PR #323, opened 2019-07-11
Native tests + Separate enums

salmamali:
Adds tests for non deprecated methods in the Android native class for the module `BugReporting`.

Adds `ArgsRegistry` class that includes all `BugReporting` enums.


android/src/test/java/com/instabug/reactlibrary/RNInstabugReactnativeModuleTest.java
```diff
@@ -69,5 +81,492 @@ public Boolean answer(InvocationOnMock invocation) throws Throwable {
...
+
+    @Test
+    public void givenArray$setReportTypes_whenQuery_thenShouldCallNativeApiWithEnumArgs() {
+        // given
+        PowerMockito.mockStatic(BugReporting.class);
+        final Map<String, Object> args = new HashMap<>();
+        ArgsRegistry.registerInstabugReportTypesArgs(args);
+        final String[] keysArray = args.keySet().toArray(new String[0]);
+
+        ReadableArray givenArray = new ReadableArray() {
```
HeshamMegid: Do we really need this? There's no way to use the actual object rather than a mock?

salmamali: ReadableArray is a react native type, and it's an abstract class so there's no way to create an instance from it. Would a better approach be to create a class that extends it for testing purposes?

PR #328, opened 2019-07-15
Use bash instead of sh in android upload_sourcemap task. Fix search for instabug token.

huanvu:
I'm not super familiar with shell scripting so feel free to correct if I'm wrong.

[[ is specific to bash so should be ran using bash instead of sh.

If INSTABUG_APP_TOKEN == "YOUR_APP_TOKEN", these next two conditionals would never run and INSTABUG_APP_TOKEN will never be found.


alyezz:
@huanvu Thanks for pointing this out, the script is fixed in 8.5.1

huanvu:
@alyezz the use of sh in build.gradle is still problematic. [[ requires bash. Could work around it by symbolically linking bash to sh but thats additional work in a CI env.

PR #344, opened 2019-07-25
e2e/ios setup

AliAbdelfattah:
Setup e2e testing with detox and jest for iOS on circle ci.


InstabugSample/ios/Podfile
```diff
@@ -23,8 +23,6 @@ target 'InstabugSample' do
   pod 'React-jsiexecutor', :path => '../node_modules/react-native/ReactCommon/jsiexecutor'
   pod 'React-jsinspector', :path => '../node_modules/react-native/ReactCommon/jsinspector'
   pod 'yoga', :path => '../node_modules/react-native/ReactCommon/yoga'
-  
-  pod 'OCMock', '~> 3.4'
```
HeshamMegid: Why is it removed from the Podfile? How is it linked to the sample app?

AliAbdelfattah: It's now linked the old way via "Linked Frameworks and Libraries".

When OCMock was being installed via pods, it was referencing the source
files instead of the pre-built binaries which contain x86_64 version
that's required when building in Release mode.

PR #346, opened 2019-07-29
🔨 refactor js code

salmamali:
- Move constants to a single shared file
- Use the class `IBGEventEmitter` to handle all JS events.


alyezz:
I think we should split `InstabugLog` into its own module like flutter

PR #354, opened 2019-08-06
Perform block fix

alyezz:
* Replace perform block with perform selector


ios/RNInstabug/InstabugReactBridge.m
```diff
@@ -360,15 +359,11 @@ - (dispatch_queue_t)methodQueue {
 }
 
 RCT_EXPORT_METHOD(setColorTheme:(IBGColorTheme)colorTheme) {
-    [[NSRunLoop mainRunLoop] performBlock:^{
```
HeshamMegid: Why is this and `setPrimaryColor` removed with no replacement?

alyezz: It works without needing to run on the main loop, i tested several cases and it seems to always work as expected, so i removed it.

PR #355, opened 2019-08-07
Fix/string mappings in iOS

AliAbdelfattah:
#### What's done:
- Fixed string mapping for addVideoMessage in iOS in accordance with the
existing Android key.
- Fixed string mapping for conversationsHeaderTitle.


alyezz:
We should be very careful when we merge salma's refactor js code to include this change in her PR.

PR #391, opened 2019-09-11
Fix android build ignoring `instabugUploadEnable` property

elyalvarado:
The previous boolean expression for checking the `instabugUploadEnable` always returned true even if the `instabugUploadEnable` ext property was set to `false`. I must confess that either I'm currently too tired, or I don't know enough to understand why this was happening.

Adding the parentheses fixes it. I manually tested all the cases:

* Not adding the `instabugUploadEnable` property: Source map is **uploaded**
* Adding the `instabugUploadEnable` property with true as value: Source map is **uploaded**
* Adding the `instabugUploadEnable` property with false as value: Source map is **not uploaded**

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -51,7 +51,7 @@ task upload_sourcemap(type: Exec) {
 
 tasks.whenTaskAdded { task ->
     if (task.name == 'preReleaseBuild' &&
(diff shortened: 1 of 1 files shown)


alyezz:
@elyalvarado Thanks!

PR #399, opened 2019-09-13
Revert android initialization from JS

alyezz:
* remove android initialization from JS side
* updates changelog
* updates readme


leethree:
Any reason for the revert?

HeshamMegid:
@leethree sorry, we missed your reply here. Unfortunately, some issues related to this slipped our testing. Since this affects the core functionality of the SDK and we still don't have an estimate for when we'd be able to fix those issues, we decided to revert the changes instead.

PR #411, opened 2019-09-26
Feature - Repro Steps

alyezz:
* Adds repro steps feature from RN side for both libraries below:

    * `react-navigation`
    * `react-native-navigation`

* Updates README


gilbertl:
Hello. We're interested in enabling repro steps in our app. Seems like this PR has been out for review since last September. What's left for this to be supported? Anyway we can help?

AliAbdelfattah:
@gilbertl thank you for offering to help :) 
Repro Steps is currently in the final stages and should be released within a couple of weeks (if not sooner).

PR #413, opened 2019-09-30
Fix relative path for ios/upload_sourcemap.sh

andreibarabas:
Currently, the Upload Sourcemap phase fails silently on iOS react native

### DIFF
--- a/ios/upload_sourcemap.sh
+++ b/ios/upload_sourcemap.sh
@@ -1,7 +1,5 @@
 #!/bin/sh
 cd ..
-cd ..
-cd ..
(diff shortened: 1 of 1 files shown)


AliAbdelfattah:
Hi @andreibarabas, could you please elaborate a bit more on why we need to update the path?

The way we do it i.e. `cd ../../..` lets us navigate from the path where the file *upload_sourcemap.sh* currently is, which is 
`{project_root}/node_modules/instabug-reactnative/ios/upload_sourcemap.sh` to `{project_root}` path, where we want the script commands to run at.

AliAbdelfattah:
Closing for inactivity.

PR #422, opened 2019-10-14
Fixing automatic soucemap upload script

alyezz:
* Removes the extra single quotations after greping the token.


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## master
+
+* Fixes automatic upload soucemap files failing to find the correct token in some cases
```
AliAbdelfattah:
```suggestion
* Fixes the automatic uploading of the source map files in some cases due to incorrect regex.
```

PR #428, opened 2019-11-05
Release 8.7.2

alyezz:
* Updates native SDKs


ios/Instabug.framework/Headers/Instabug.h
```diff
@@ -321,6 +321,12 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (class, atomic, assign) IBGSDKDebugLogsLevel sdkDebugLogsLevel;
 
+
+#pragma mark - Private Methods for CrossPlatform   /// Will  be removed before the release
```
HeshamMegid: These shouldn't be in the header. Can you please check with iOS about this?

cc: @Kmohamed

InstabugSample/__tests__/instabugUtils.spec.js
```diff
@@ -56,25 +56,25 @@ describe('Test global error handler', () => {
 
     });
 
-    it('should emit event IBGSendUnhandledJSCrash when platform is android and onReportSubmitHandler is set', (done) => {
```
HeshamMegid: Why are we removing this?

ios/Instabug.framework/Headers/Instabug.h
Kmohamed: @HeshamMegid we will remove it in this release

InstabugSample/__tests__/instabugUtils.spec.js
alyezz: This is the same test we commented in a diffferent PR, the one that broke even on previous passing builds. Seems to be a problem with the test

PR #488, opened 2020-07-01
Fix Android initialization with no invocation events

romanbb:
Null check the array before accessing it.

### DIFF
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativePackage.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativePackage.java
@@ -73,21 +73,22 @@ public class RNInstabugReactnativePackage implements ReactPackage {
     }
 
     private void parseInvocationEvent(String[] invocationEventValues) {
+        if (invocationEventValues != null) {
+            for (int i = 0; i < invocationEventValues.length; i++) {
+                if (invocationEventValues[i].equals("button")) {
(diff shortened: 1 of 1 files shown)


AliAbdelfattah:
Hi @romanbb, thank you for submitting a PR!

Would you mind sharing the use case where the invocation events need to be passed as null? What is it you want to achieve but can't with the current implementation?

romanbb:
> Hi @romanbb, thank you for submitting a PR!
> 
> Would you mind sharing the use case where the invocation events need to be passed as null? What is it you want to achieve but can't with the current implementation?

Hey @AliAbdelfattah - sure!

Our use case was that we are including the Instabug library via the RN package. We did not want to load the library for development builds. So we simply omitted the following call when it was a development build:

    Instabug.startWithToken(<>, [Instabug.invocationEvent.shake])

And this produced the crash. A third party library shouldn't be causing a crash if not invoked I think. Perhaps a warning, or a more useful exception can be thrown if you feel strongly about the usage of the API and requiring a `startWithTokenCall()`.

TheBuggedYRN:
Thanks @romanbb for this PR!
But since v11, the SDK can now be initialized from JS only, you do not need handle this from your Java code anymore.

Here's what you can do for now:

1. Check out [v11 migration guide](https://docs.instabug.com/docs/react-native-migration-guide) and move your code to JS.
1. Consider using `Instabug.invocationEvent.none` instead of null or an empty array.

PR #528, opened 2020-10-12
Fixing Yarn issue in CI

alyezz:
* Fixes an issue causing the CI to fail due to Yarn

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,7 @@ jobs:
     steps:
       - checkout:
           path: ~/project
-      - run: yarn install
+      - run: yarn --ignore-engines
       - run:
           name: chmod permissions
           command: cd android && chmod +x ./gradlew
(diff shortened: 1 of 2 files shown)


AliAbdelfattah (changes requested):
It's probably not the best idea to ignore the node engine version incompatibilities. Although it's working now, it could easily break in the future.

Having a look at the release history for the library `teeny-request` which `codecov` depends on, I can see in v7 they've upgraded their node engine. Updating `codecov` from v3.7 to v3.8 which in turn updates `teeny-request` from v6 to v7 which causes incompatibility.

A temporary solution [here](https://github.com/Instabug/Instabug-React-Native/commit/2f902dd61576cd4ba66bbb1dbd694d65ed31ef1a) until we decide to upgrade our CI node engine.

AliAbdelfattah (approved):

PR #556, opened 2021-03-02
Fix discrepancies reported by the type validator

ahmed1hisham:
- Fixes discrepancies reported by the type validator, hence, resolves CircleCI errors.

### DIFF
--- a/index.js
+++ b/index.js
@@ -646,11 +646,11 @@ const InstabugModule = {
    * @deprecated use {@link CrashReporting.reportJSException}
    * Send handled JS error object
    *
-   * @param Exception   Error object to be sent to Instabug's servers
+   * @param errorObject Error object to be sent to Instabug's servers
    */
 
-  reportJSException(Exception) {
-    CrashReporting.reportJSException(Exception);
+  reportJSException(errorObject) {
+    CrashReporting.reportJSException(errorObject);
   },
 
   /**
--- a/index.js
+++ b/index.js
@@ -251,10 +251,10 @@ const InstabugModule = {
    * Sets the color of UI elements indicating interactivity or call to action.
    * To use, import processColor and pass to it with argument the color hex
    * as argument.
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -646,11 +646,11 @@ const InstabugModule = {
...
    * Send handled JS error object
    *
-   * @param errorObject   Error object to be sent to Instabug's servers
+   * @param Exception   Error object to be sent to Instabug's servers
    */
 
-  reportJSException(errorObject) {
-    CrashReporting.reportJSException(errorObject);
+  reportJSException(Exception) {
+    CrashReporting.reportJSException(Exception);
```
AliAbdelfattah, line 653: Although Exception and Error are both synonymous, it's the Error term that dominates in JS. Can we update the TS file instead please?

modules/CrashReporting.js
```diff
@@ -21,12 +21,12 @@ export default {
...
-   * @param errorObject   Error object to be sent to Instabug's servers
+   * @param Exception   Error object to be sent to Instabug's servers
    */
-  reportJSException: function(errorObject) {
-    let jsStackTrace = InstabugUtils.parseErrorStack(errorObject);
+  reportJSException: function(Exception) {
+    let jsStackTrace = InstabugUtils.parseErrorStack(Exception);
     var jsonObject = {
-      message: errorObject.name + ' - ' + errorObject.message,
+      message: Exception.name + ' - ' + Exception.message,
```
AliAbdelfattah, line 29: Same here.

modules/CrashReporting.js
```diff
@@ -21,15 +21,15 @@ export default {
   /**
    * Send handled JS error object
    *
-   * @param errorObject Error object to be sent to Instabug's servers
+   * @param Exception   Error object to be sent to Instabug's servers
```
AliAbdelfattah, line 24: Should we revert this one?

PR #563, opened 2021-04-01
[INSD-5139] Fix handling network request body that is not a string

ahmed1hisham:
## Description of the change
- Added a type check on the request body of a network log in the `NetworkLogger.js` module
- Stringified the request body in case it's not a string before passing it to the native layer.
- Changes done for iOS only
## Type of change
### Development
- [ ] Lint rules pass locally
### Code review 

### DIFF
--- a/modules/NetworkLogger.js
+++ b/modules/NetworkLogger.js
@@ -33,9 +33,6 @@ export default {
               if (Platform.OS === 'android') {
                 Instabug.networkLog(JSON.stringify(network));
               } else {
-                if (typeof network.requestBody !== "string") {
-                  network.requestBody = JSON.stringify(network.requestBody);
-                }
                 Instabug.networkLog(network);
               }
             } catch (e) {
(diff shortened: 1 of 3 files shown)


modules/NetworkLogger.js
```diff
@@ -33,6 +33,9 @@ export default {
               if (Platform.OS === 'android') {
                 Instabug.networkLog(JSON.stringify(network));
               } else {
+                if (typeof network.requestBody !== "string") {
+                  network.requestBody = JSON.stringify(network.requestBody);
```
AliAbdelfattah, line 37: It's best to keep the data manipulation in one place. Could you move the formatting to the Xhr 
interceptor please?

AliAbdelfattah (changes requested):

modules/NetworkLogger.js
ahmed1hisham: Yes, sure.

PR #578, opened 2021-04-29
[MOB-5167] Fix/android sourcemap upload

ahmed1hisham:
## Description of the change
- Fixed Hermes sourcemap generation for a successful upload process
## Type of change
### Development
- [ ] Lint rules pass locally
### Code review 

### DIFF
--- a/android/upload_sourcemap.sh
+++ b/android/upload_sourcemap.sh
@@ -39,6 +39,23 @@ else
     echo "Instabug: Generating sourcemap files..."
     IS_HERMES=$(grep "enableHermes:" ./android/app/build.gradle -m 1)
     if [[ $IS_HERMES == *"true"* ]]; then
+        #Find HERMES OS bin directory
+        case "$OSTYPE" in
+            darwin*)  HERMES_OS_BIN='osx-bin' ;; 
+            linux*)   HERMES_OS_BIN='linux64-bin' ;;
+            msys*)    HERMES_OS_BIN='win64-bin' ;;
+            *)        echo "unknown: $OSTYPE" ;;
+        esac
+
+        #Find HERMES command file name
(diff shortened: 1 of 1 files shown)


android/upload_sourcemap.sh
```diff
@@ -39,6 +39,25 @@ else
...
+        #Find HERMES OS bin directory
+        case "$OSTYPE" in
+            darwin*)  HERMES_OS_BIN='osx-bin' ;; 
+            linux*)   HERMES_OS_BIN='linux64-bin' ;;
+            msys*)    HERMES_OS_BIN='win64-bin' ;;
+            *)        echo "unknown: $OSTYPE" ;;
+        esac
+
+        #Find HERMES command file name
+        RN_VERSION=$(npm view react-native version)
```
AliAbdelfattah, line 51: I believe this command gets the most recent version regardless of the one installed. I was thinking of a more direct approach. What do you think?
```
INSTALLED_RN_VERSION=`node -p "require('<pkg_path_in_node_modules>/package.json').version"`
```

android/upload_sourcemap.sh
```diff
@@ -39,6 +39,25 @@ else
...
+        VERSION_NO=$( echo $RN_VERSION | cut -d "." -f2 )
```
AliAbdelfattah, line 52: Let's clarify it a bit more since version and version number are both synonyms to the same thing. How about INSTALLED_RN_VERSION_MAJOR?

AliAbdelfattah (changes requested):

android/upload_sourcemap.sh
ahmed1hisham: True, that command is more accurate. Thanks, I'll use it.

android/upload_sourcemap.sh
ahmed1hisham: Makes sense, will do.

AliAbdelfattah (approved):

PR #625, opened 2021-09-24
Fix android/upload_sourcemap.sh

gilbertl:
## Description of the change
* use bash shell to run the script, otherwise it'll fail with "[[: not found"
* see https://stackoverflow.com/a/50957629 for SO explanation

## Type of change
- [x ] Bug fix (non-breaking change that fixes an issue)

### Development
- [x ] Lint rules pass locally
- [] The code changed/added as part of this pull request has been covered with tests
### Code review 


gilbertl:
This is for our android app, and I think CircleCi only has android build machines on Linux. 

If even if they do support MacOS, I know those machines are more expensive.

PR #675, opened 2022-03-07
[MOB-9131] Symlink the SDK in the sample app

a7med-mahmoud:
## Description of the change

Symlinks the Instabug SDK in the sample app's `node_modules` instead of copying the whole Instabug-React-Native repo files again in the `node_modules` folder.

## Benefits

- Update SDK files on the fly (no need to `yarn install` every time you do a change) with hot reload when running the sample app.
- Update native SDK code in the sample app's Android and iOS projects — This makes autosuggestions in Android Studio and Xcode much better.
- More efficient storage usage — No need to copy the whole repo each time you install, it also prevents recursive file copying when doing multiple installs.

## Type of change
### Development
- [ ] Lint rules pass locally
### Code review 

### DIFF
--- a/InstabugSample/metro.config.js
+++ b/InstabugSample/metro.config.js
@@ -5,7 +5,15 @@
  * @format
  */
 
+const path = require('path');
+const escape = require('escape-string-regexp');
+const exclusionList = require('metro-config/src/defaults/exclusionList');
+
+const root = path.resolve(__dirname, '..');
+const modules = ['react', 'react-native'];
+
 module.exports = {
+  watchFolders: [root],
   transformer: {
     getTransformOptions: async () => ({
       transform: {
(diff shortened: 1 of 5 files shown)


InstabugSample/metro.config.js
```diff
@@ -5,7 +5,15 @@
  * @format
  */
 
+const path = require('path');
+const escape = require('escape-string-regexp');
+const exclusionList = require('metro-config/src/defaults/exclusionList');
+
+const root = path.resolve(__dirname, '..');
+const modules = ['react', 'react-native', '@babel/runtime'];
```
TheBuggedYRN, line 13: IMO, extracting `react` and `react-native` from `peerDependencies` will decouple this file from the changes made to the SDK's `package.json`.

a7med-mahmoud: Good call, but the thing is that we don't actually have any `peerDependencies` in the SDK's `package.json` so I added them as an array in the `metro.config.js` file, another solution would be to add them as `peerDependencies` and get them dynamically in `metro.config.js`.

TheBuggedYRN: We can add them for now with unconstrained versions `"*"`.

DavidMina96 (approved):

AliAbdelfattah (approved):

TheBuggedYRN (approved):

PR #693, opened 2022-04-15
Allow to skip uploading sourcemaps with env variable INSTABUG_SKIP_UPLOAD

christophehurpeau:
## Description of the change

For faster tests builds where we don't need to send source maps, this variable `INSTABUG_SKIP_UPLOAD` allows us to skip this part

## Type of change


ios/upload_sourcemap.sh
```diff
@@ -30,6 +30,11 @@ if [ ! "${INSTABUG_APP_TOKEN}" ] || [ -z "${INSTABUG_APP_TOKEN}" ]; then
     echo "Instabug: INSTABUG_APP_TOKEN not found. Make sure you've added the SDK initialization line Instabug.start Or added the environment variable INSTABUG_APP_TOKEN"
     exit 0
 else
+    if [[ $INSTABUG_SKIP_UPLOAD == "true" ]]; then
```
christophehurpeau, line 33:
```suggestion
    if [[ "${INSTABUG_SKIP_UPLOAD}" == "true" ]]; then
```

PR #720, opened 2022-06-09
[MOB-8883] Add CI job to validate shell files and increase e2e_android resources

AliAbdelfattah:
## Description of the change
- Add CI job to validate shell files
- Fix Android shell script error
- Fix e2e_android flaky runs due to low resources

## Type of change
- [X] Bug fix (non-breaking change that fixes an issue)

### Code review 
- [X] This pull request has a descriptive title and information useful to a reviewer
- [X] Issue from task tracker has a link to this pull request

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -163,6 +163,7 @@ jobs:
     executor:
       name: android/android-machine
       tag: "2022.03.1"
+      resource-class: large
     working_directory: ~/project/InstabugSample/android
     steps:
       - checkout:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -89,6 +89,21 @@ jobs:
           working-directory: ./android
           test-command: ./gradlew test
 
+  validate_shell_files:
+    executor:
+      name: android/android-machine
+      tag: "2022.03.1"
```
TheBuggedYRN, line 95: We can run it on a linux machine instead? What do you think? That would be relatively faster.

TheBuggedYRN (changes requested):

PR #788, opened 2022-10-03
[INSD-8104] Declare local vars final

DavidMina96:
### Problem:
Users face a compile error on Android stating that `local variables accessed from within inner class need to be declared final`. This is only encountered on projects with Java version prior to 8. 

**Starting Java 8**, this requirement was relaxed to final or _effectively final_. A variable whose value is never changed after it is initialized is effectively final. Hence, users on Java 8 or higher do not face this issue. 

### Solution:
Declare these local variables _explicitly_ `final` so that the app builds successfully on different Java versions. 
## Type of change
### Development
- [ ] Lint rules pass locally
### Code review 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * Uses Cocoapods for Instabug iOS SDK
+* Fixes a compile error on projects with Java version prior to 8.
 
 ## 11.2.0 (2022-09-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Unreleased
 
 * Uses Cocoapods for Instabug iOS SDK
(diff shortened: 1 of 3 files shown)


TheBuggedYRN (dismissed):

a7med-mahmoud (changes requested):
I think also we need to do the same thing to ```RNInstabugReactnativeModule.setInvocationEvents``` parameters.

DavidMina96:
@a7med-mahmoud The compiler didn't complain about it, but yes we can better do the same here.

a7med-mahmoud (dismissed):

TheBuggedYRN (dismissed):

a7med-mahmoud (approved):

TheBuggedYRN (approved):

AliAbdelfattah (approved):

PR #852, opened 2022-12-13
[INSD-8643] Add `reportError` and deprecate `reportJSException`

DavidMina96:
## Description of the change

**Problem:**
Enforcing a strong type for `reportJSException` would require code changes from the users as the `catch` block parameter has to be of type `any` / `unknown`.

**Solution:**
1. Instead of the users having to handle the `error` type checking before calling the API, we accept `any` type and handle the type checking internally not to cause breaking changes with the current API.
2. Introduce a new strongly typed API `reportError` and deprecate `reportJSException`

## Type of change

## Related issues

> Issue links go here

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/modules/CrashReporting.ts
+++ b/src/modules/CrashReporting.ts
@@ -1,5 +1,4 @@
 import { Platform } from 'react-native';
-import type { ExtendedError } from 'react-native/Libraries/Core/Devtools/parseErrorStack';
 
 import { NativeCrashReporting } from '../native';
 import IBGEventEmitter from '../utils/IBGEventEmitter';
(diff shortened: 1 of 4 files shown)


a7medev (approved):

src/modules/CrashReporting.ts
```diff
@@ -16,10 +16,21 @@ export const setEnabled = (isEnabled: boolean) => {
...
+  if (error instanceof Error) {
+    reportError(error);
+  }
+};
+
+/**
+ * Send handled JS error object
+ * @param error Error object to be sent to Instabug's servers
+ */
+export const reportError = (error: ExtendedError) => {
```
TheBuggedYRN, line 33: **suggestion:** Let's use the new API in the example app instead of the old one; to demonstrate how it should be used.

TheBuggedYRN (changes requested):

PR #867, opened 2022-12-28
[MOB-11439] Add `init` API

DavidMina96:
## Description of the change

We need our SDK initialization API to take an extensible configuration rather than a fixed set of parameters. For that, we add a new initialization API `Instabug.init` and deprecate `Instabug.start`. 

This PR includes:

- Adding a new API `Instabug.init` that takes an `InstabugConfig`
- Adding `init` API tests
- Deprecating `start` and removing its mappings from the native bridges, along with its tests
- Using `init` instead of `start` in example app
- Updating the source maps scripts to get the app token from the new `init` API
- Updating README and CHANGELOG

## Type of change

## Related issues

> Issue links go here

### Development

- [x] Lint rules pass locally

### Code review


android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
```diff
@@ -127,15 +127,13 @@ public void run() {
...
-   */
-    @ReactMethod
-    public void start(final String token, final ReadableArray invocationEventValues) {
+    /**
+     * Initializes the SDK.
+     * @param token The token that identifies the app. You can find it on your dashboard.
+     * @param invocationEventValues The events that invoke the SDK's UI.
+     */
+    @ReactMethod
+    public void init(final String token, final ReadableArray invocationEventValues) {
```
a7medev, line 136: It think we should make it take a `ReadableMap options` argument just like the JavaScript API to prevent the number of parameters from growing and being unreadable as we add more options. What do you think?

ios/RNInstabug/InstabugReactBridge.m
```diff
@@ -41,8 +41,8 @@ - (dispatch_queue_t)methodQueue {
     Instabug.enabled = isEnabled;
 }
 
-RCT_EXPORT_METHOD(start:(NSString *)token invocationEvents:(NSArray*)invocationEventsArray) {
-     SEL setPrivateApiSEL = NSSelectorFromString(@"setCurrentPlatform:");
+RCT_EXPORT_METHOD(init:(NSString *)token invocationEvents:(NSArray*)invocationEventsArray) {
```
a7medev, line 44: _Same comment as the one on Java method._

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
TheBuggedYRN: I think we will lose the type-safety if we went with that approach.

tests/modules/Instabug.spec.js
```diff
@@ -136,20 +136,22 @@ describe('Instabug Module', () => {
...
+    expect(NativeInstabug.init).toBeCalledWith(
+      instabugConfig.token,
+      instabugConfig.invocationEvents,
+    );
   });
 
-  it('should report the first screen on SDK start', async () => {
-    Instabug.start('some-token');
+  it('should report the first screen on SDK initialization', async () => {
+    Instabug.init('some-token');
```
TheBuggedYRN, line 154: **issue:** This line is possibly not properly migrated.

```suggestion
    Instabug.init({ token: 'some-token' });
```

tests/mocks/mockInstabug.js
```diff
@@ -2,6 +2,7 @@ export default {
   Instabug: {
     setEnabled: jest.fn(),
     start: jest.fn(),
```
TheBuggedYRN, line 4: **suggestion:** We can remove `start` as it's no longer in the native bridge.

src/modules/Instabug.ts
```diff
@@ -46,6 +46,11 @@ export {
   strings,
 };
 
+interface InstabugConfig {
```
TheBuggedYRN, line 49: **suggestion:** I think we should export that type to the users, what do you think?

tests/modules/Instabug.spec.js
```diff
@@ -38,7 +38,7 @@ describe('Instabug Module', () => {
   });
 
   it("componentDidAppearListener shouldn't call the native method reportScreenChange if first screen", async () => {
-    Instabug.start('some-token');
+    Instabug.init('some-token');
```
TheBuggedYRN, line 41: **issue:** This line is possibly not properly migrated.

```suggestion
    Instabug.init({ token: 'some-token' });
```

TheBuggedYRN (changes requested):

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
DavidMina96: @a7medev I second Youssef on that.

src/modules/Instabug.ts
DavidMina96: Makes sense!

TheBuggedYRN (approved):

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
a7medev: @DavidMina96 @TheBuggedYRN `ReadableMap` has methods to get data with while specifying the type like `getInt`, `getBoolean`, etc; which can help in type-safety, we won't lose type-safety completely.

TheBuggedYRN: @a7medev Having to implicitly cast the `Object`s inside the `ReadableMap` using the `getSomeType` would be still be less type-safe than the explicit parameters, IMO. Because, not only we will be compensating on the types, but we will have to make sure that the magic strings used for keys (e.g. `"token"`) still stay in sync. If the new architecture has a solution for both of these problems, we can go with that change by then.

a7medev: Sounds reasonable, let's make keep it as is then.

a7medev (approved):

TheBuggedYRN (approved):

PR #877, opened 2023-01-02
[MOB-11441] Add CodePush Support

DavidMina96:
## Description of the change

- Add `codePushLabel` property to `InstabugConfig` and its native mappings on both platforms, along with the needed tests.

## Type of change

## Related issues

> Issue links go here

### Development

- [x] Lint rules pass locally

### Code review


example/ios/InstabugSampleTests/InstabugSampleTests.m
```diff
@@ -218,6 +219,15 @@ - (void)testSetReproStepsMode {
   OCMVerify([mock setReproStepsMode:reproStepsMode]);
 }
 
+- (void)testSetCodePushVersion {
+  id mock = OCMClassMock([Instabug class]);
+   NSString *codePushLabel = @"v123";
+
+   OCMStub([mock setCodePushVersion:codePushLabel]);
+   [self.instabugBridge setCodePushVersion:codePushLabel];
+   OCMVerify([mock setCodePushVersion:codePushLabel]);
```
a7medev, line 228: Small formatting suggestion to keep the spacing consistent.

```suggestion
  NSString *codePushLabel = @"v123";

  OCMStub([mock setCodePushVersion:codePushLabel]);
  [self.instabugBridge setCodePushVersion:codePushLabel];
  OCMVerify([mock setCodePushVersion:codePushLabel]);
```

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
```diff
@@ -134,7 +134,7 @@ public void run() {
      * @param invocationEventValues The events that invoke the SDK's UI.
      */
     @ReactMethod
-    public void init(final String token, final ReadableArray invocationEventValues, final String logLevel) {
+    public void init(final String token, final ReadableArray invocationEventValues, final String codePushLabel, final String logLevel) {
```
a7medev, line 137: I wonder where the word `label` came from as we already use `version` on both iOS and Android, is there a reason why it's not just `codePushVersion`?

example/ios/InstabugSampleTests/InstabugSampleTests.m
```diff
@@ -66,12 +66,13 @@ - (void)testInit {
...
   IBGSDKDebugLogsLevel sdkDebugLogsLevel = IBGSDKDebugLogsLevelDebug;
   
   XCTestExpectation *expectation = [self expectationWithDescription:@"Testing [Instabug init]"];
   
   OCMStub([mock startWithToken:appToken invocationEvents:floatingButtonInvocationEvent]);
-  [self.instabugBridge init:appToken invocationEvents:invocationEvents debugLogsLevel:sdkDebugLogsLevel];
+  [self.instabugBridge init:appToken invocationEvents:invocationEvents codePushLabel:codePushLabel debugLogsLevel:sdkDebugLogsLevel];
 
   [[NSRunLoop mainRunLoop] performBlock:^{
     OCMVerify([mock startWithToken:appToken invocationEvents:floatingButtonInvocationEvent]);
```
a7medev, line 78: I think we also need to verify that the `setCodePushVersion:` has been called with the correct `codePushLabel` in the `init` test to ensure that it does indeed call it.

tests/modules/Instabug.spec.js
```diff
@@ -149,6 +150,7 @@ describe('Instabug Module', () => {
     expect(NativeInstabug.init).toBeCalledWith(
       instabugConfig.token,
       instabugConfig.invocationEvents,
+      instabugConfig.codePushLabel,
```
a7medev, line 153: I think we should also test the case when the `codePushLabel` is not passed and verify that it was called with an empty string `''`.

DavidMina96: Makes sense! I added another test for `init` to verify the native method is called with the default values given the optional properties aren't passed.

a7medev (approved):
LGTM 👏

PR #879, opened 2023-01-05
[MOB-9960] Create CLI for Source Maps Upload

DavidMina96:
## Description of the change

Create a TS CLI with Node and Commander that supports uploading source maps via `upload-sourcemaps` command.

## Type of change

## Related issues

> Issue links go here

### Development

- [x] Lint rules pass locally

### Code review


cli/sourcemaps.ts
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env node
+import { Command, Option, OptionValues } from 'commander';
+import figlet from 'figlet';
+import fs from 'fs';
+
+const program: Command = new Command();
+
+console.log(figlet.textSync('Instabug React Native CLI'));
```
TheBuggedYRN, line 8: **suggestion:** I think "Instabug CLI" is less verbose and matches the usage of `yarn instabug upload-sourcemaps`. What do you think?

```suggestion
console.log(figlet.textSync('Instabug CLI'));
```

cli/sourcemaps.ts
```diff
@@ -0,0 +1,64 @@
...
+  .addOption(
+    new Option('-c, --code <value>', 'The app version code')
+      .env('INSTABUG_APP_VERSION_CODE')
+      .makeOptionMandatory(),
+  )
+  .option('-l, --label <value>', "The CodePush label if it's a CodePush release")
+  .showHelpAfterError()
+  .parse(process.argv);
+
+const options: OptionValues = program.opts();
```
TheBuggedYRN, line 40: **suggestion:** We can make this more type safe as follows:

```suggestion
interface UploadSourcemapsOptions {
  platform: 'android' | 'ios';
  dir: string;
  token: string;
  name: string;
  code: number;
  label: string;
}

const options = program.opts<UploadSourcemapsOptions>();
```

tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+  "exclude": ["cli/**/*"],
```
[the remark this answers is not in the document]
DavidMina96: Yes, it will yield the same result; I agree it's better to be more explicit about what we really want to **include**. 👍

package.json
```diff
@@ -27,7 +30,8 @@
     "format": "prettier --check . --ignore-path .gitignore",
     "format:fix": "prettier --write . --ignore-path .gitignore",
     "test": "jest",
-    "build": "tsc"
+    "build": "tsc",
+    "build:cli": "tsc --p tsconfig.cli.json"
```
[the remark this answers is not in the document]
DavidMina96: I don't see a clear benefit, but it looks nicer to me. Let's have it the way you suggested. 👍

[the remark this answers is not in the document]
DavidMina96: The `build` script will have to be `"build": "yarn build:lib && yarn build:cli"` though in this case.

package.json
```diff
@@ -10,6 +10,9 @@
   "main": "dist/index",
   "types": "dist/index.d.ts",
   "react-native": "src/index.ts",
+  "bin": {
+    "upload-sourcemaps": "bin/sourcemaps.js"
```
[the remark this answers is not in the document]
DavidMina96: Nice catch! I recall we agreed on the command usage to be as follows: `yarn instabug-reactnative upload-sourcemaps` so I think we should go for naming the binary `instabug-reactnative` and discuss it further internally.

cli/sourcemaps.ts
```diff
@@ -0,0 +1,64 @@
...
+    new Option('-v, --name <value>', 'The app version name')
+      .env('INSTABUG_APP_VERSION_NAME')
+      .makeOptionMandatory(),
+  )
+  .addOption(
+    new Option('-c, --code <value>', 'The app version code')
+      .env('INSTABUG_APP_VERSION_CODE')
+      .makeOptionMandatory(),
+  )
+  .option('-l, --label <value>', "The CodePush label if it's a CodePush release")
```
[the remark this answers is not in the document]
DavidMina96: We should stick with `addOption` then since we need to construct an `Option` object to handle some extra configuration (e.g. env, choices) for some options.

cli/sourcemaps.ts
DavidMina96: It depends on the command usage as you mentioned, which we'll discuss.

cli/sourcemaps.ts
DavidMina96: Good suggestion! I just prefer a `code` type of `string` because it'll be sent to the endpoint as a string at the end.

cli/UploadSourcemaps.ts
```diff
@@ -0,0 +1,88 @@
...
+      .makeOptionMandatory(),
+  )
+  .addOption(new Option('-l, --label <value>', "The CodePush label if it's a CodePush release"))
+  .action(function (this: Command) {
+    const options = this.opts<UploadSourcemapsOptions>();
+    uploadFile(options);
+  })
+  .showHelpAfterError();
+
+interface UploadSourcemapsOptions {
```
TheBuggedYRN, line 44: **suggestion:** Let move the interface definition to the top of the file, right after the imports, since it's used in previous lines of code.

cli/UploadSourcemaps.ts
```diff
@@ -0,0 +1,88 @@
...
+  const form = new FormData();
+  form.append('app_version', JSON.stringify(version));
+  form.append(
+    'symbols_file',
+    fs.readFileSync(path.join(dir, `${os}-sourcemap.json`)),
+    `${os}-sourcemap.json`,
+  );
+  form.append('application_token', token);
+  form.append('platform', 'react_native');
+  form.append('os', os);
```
TheBuggedYRN, line 74: **suggestion:** This might contains multiple comments, just to show the final suggestion:

1. A more descriptive function name can be `uploadFile`.
1. It might be a matter of taste, but I think the re-assigning of `options` might be pretty verbose. I would suggest using `opts` for short and more readable inline code. What do you think?
1. Using blank lines between related lines of code can improve readability.
1. Extracting varaibles related to variables might be more readable as well IMO, instead of having a compound line. A rule of thumb I use is _"Never call a function inside the argument of another function"_. 

```suggestion
const uploadSourcemaps = async (opts: UploadSourcemapsOptions) => {
  const fileName = `${opts.platform}-sourcemap.json`;
  const filePath = path.join(opts.dir, fileName);
  const fileBlob = fs.readFileSync(filePath);

  const version = {
    code: opts.code,
    name: opts.name,
    codePush: opts.code,
  };

  const form = new FormData();
  form.append('platform', 'react_native');
  form.append('application_token', opts.token);
  form.append('os', opts.platform);
  form.append('symbols_file', fileBlob, fileName);
  form.append('app_version', JSON.stringify(version));
```

cli/UploadSourcemaps.ts
```diff
@@ -0,0 +1,88 @@
...
+    new Option('-v, --name <value>', 'The app version name')
+      .env('INSTABUG_APP_VERSION_NAME')
+      .makeOptionMandatory(),
+  )
+  .addOption(
+    new Option('-c, --code <value>', 'The app version code')
+      .env('INSTABUG_APP_VERSION_CODE')
+      .makeOptionMandatory(),
+  )
+  .addOption(new Option('-l, --label <value>', "The CodePush label if it's a CodePush release"))
```
TheBuggedYRN, line 37: **question:** Should we add env `INSTABUG_APP_VERSION_LABEL` for parity with other version-related options?

cli/UploadSourcemaps.ts
```diff
@@ -0,0 +1,88 @@
...
+  );
+  form.append('application_token', token);
+  form.append('platform', 'react_native');
+  form.append('os', os);
+  console.log('Uploading Sourcemap file...');
+  try {
+    const response = await axios.post('https://api.instabug.com/api/sdk/v3/symbols_files', form, {
+      headers: {
+        ...form.getHeaders(),
+      },
```
TheBuggedYRN, line 80: **suggestion:** As long as we do not have extra headers, no need to do the destruction.
```suggestion
      headers: form.getHeaders(),
```

cli/UploadSourcemaps.ts
```diff
@@ -0,0 +1,88 @@
...
+    });
+    console.log('Success:', response.data);
```
TheBuggedYRN, line 82: **suggestion:** Let's use a more descriptive confirmation message, since the one returned by the server is pretty ambigous.
```suggestion
    console.log(`Successfully uploaded Sourcemaps for version: ${...}`);
```

cli/UploadSourcemaps.ts
```diff
@@ -0,0 +1,88 @@
...
+  } catch (error) {
+    if (axios.isAxiosError(error)) {
+      console.error('Error:', error.response?.data);
+    }
```
TheBuggedYRN, line 86: **suggestion:** I think we should log all type of errors, not just axios errors, to have more clarity about what went wrong.

cli/UploadSourcemaps.ts
```diff
@@ -0,0 +1,88 @@
...
+  form.append('app_version', JSON.stringify(version));
+  form.append(
+    'symbols_file',
+    fs.readFileSync(path.join(dir, `${os}-sourcemap.json`)),
+    `${os}-sourcemap.json`,
+  );
+  form.append('application_token', token);
+  form.append('platform', 'react_native');
+  form.append('os', os);
+  console.log('Uploading Sourcemap file...');
```
TheBuggedYRN, line 75: **suggestion (non-blocking):** Let's have a global logging prefix whenever we log something to the console, what do you think? 

```js
// cli/util.ts
export const LOG_PREFIX = "[Instabug]";
```

```suggestion
  console.log(LOG_PREFIX, "Uploading Sourcemaps...");
```

cli/UploadSourcemaps.ts
```diff
@@ -0,0 +1,88 @@
+import axios from 'axios';
+import { Command, Option } from 'commander';
+import FormData from 'form-data';
+import fs from 'fs';
+import path from 'path';
+
+export const uploadSourcemapsCommand: Command = new Command();
```
TheBuggedYRN, line 7: **suggestion:** We can remove the type annotation, since it's already interfered.

```suggestion
export const uploadSourcemapsCommand = new Command();
```

TheBuggedYRN (changes requested):

cli/index.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { uploadSourcemapsCommand } from './UploadSourcemaps';
+
+const program: Command = new Command();
+
+console.log(figlet.textSync('Instabug CLI'));
+
+program
+  .name('instaubg')
+  .version('1.0.0-beta1')
```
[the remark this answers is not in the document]
DavidMina96: I did follow the conventions [here](https://commons.apache.org/releases/versioning.html) - under Beta Release Numbers section.

cli/UploadSourcemaps.ts
DavidMina96: I agree that all the aforementioned points positively impact readability. 👏 💡

cli/UploadSourcemaps.ts
DavidMina96: I prefer we log both since we might need to know the server response status.

cli/UploadSourcemaps.ts
DavidMina96: Makes sense.

cli/UploadSourcemaps.ts
DavidMina96: I think this was done in the shell scripts just to be able to quickly spot/extract Instabug-related logs as the scripts run between lots of other phases, so I don't find it adding value in the CLI case. Let me know if you have a certain use case in mind.

cli/UploadSourcemaps.ts
DavidMina96: Yes, it's a good idea.

TheBuggedYRN (approved):

PR #931, opened 2023-02-27
[MOB-12026] Fix TS Compile Error

DavidMina96:
## Description of the change
**Problem**
A TS compile error started to arise on `v11.9.0` due to the `dist` dir having a different structure than the entry points provided in `package.json`. In particular, adding the `test` dir in `tsconfig.json` `include` property is the root cause of the change. This was done during the unit tests migration to TS to resolve an error related to the declaration file for module `xhr2`, which is inside the `test` dir. 

**Solution**
Remove the `test` dir from being included and add a separate `tsconfig.test.json` for the tests configuration, which would preserve the old `dist` structure and provide the required files for the tests to run. 

Note:
Changelog formatting according to the new guidelines will be done entirely in a separate PR.  

## Type of change

## Related issues

> Issue links go here

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "include": ["src/**/*", "test/**/*"],
+  "include": ["src/**/*"],
   "compilerOptions": {
     "module": "esnext",
     "target": "esnext",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## Unreleased
+
+- Fixes a TS compilation error due to a broken entry point path.
+
 ## 11.9.0 (2023-02-20)
 
 - Bumps Instabug Android SDK to v11.9.0.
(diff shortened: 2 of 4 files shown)


a7medev (approved):
LGTM 👏

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
-  "include": ["src/**/*", "test/**/*"],
+  "files": ["test/xhr2.d.ts"],
```
TheBuggedYRN, line 2: **suggestion:** I don't think keeping both src config and test config in one `tsconfig.json` should still be our approach here. Editing this tsconfig for the sake of tests might cause similar unwanted behaviour in the future.

I suggest moving this to `tsconfig.test.json`:

```json
{
  "extends": "./tsconfig.json",
  "include": ["test/**/*"]
}
```

Then updating `jest.config.js` to use the new `tsconfig.test.json` file.

TheBuggedYRN (changes requested):

tsconfig.json
DavidMina96: @TheBuggedYRN That's actually a good suggestion. Let's do it this way. 👍

TheBuggedYRN (approved):

PR #1002, opened 2023-07-19
[MOB-12640] Unify Repro Steps List Item Title String

a7medev:
## Description of the change

- Deprecate `StringKey.reproStepsListItemTitle` as it has inconsistent behavior between Android and iOS (it puts an extra ` #` on iOS and doesn't put it on Android).
- Introduce a consistent `StringKey.reproStepsListItemNumberingTitle` which adds the ` #` on Android as well.

## Type of change

## Related issues

> Issue links go here

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/native/NativeConstants.ts
+++ b/src/native/NativeConstants.ts
@@ -176,7 +176,7 @@ interface NativeStringKey {
   reproStepsListEmptyStateDescription: any;
   reproStepsListHeader: any;
   reproStepsListItemTitle: any;
-  reproStepsListItemNumberingTitle: string;
+  reproStepsListItemNumberingTitle: any;
   reproStepsProgressDialogBody: any;
   requestFeatureDescription: any;
   screenRecording: any;
--- a/src/native/NativeConstants.ts
+++ b/src/native/NativeConstants.ts
@@ -176,6 +176,7 @@ interface NativeStringKey {
   reproStepsListEmptyStateDescription: any;
   reproStepsListHeader: any;
   reproStepsListItemTitle: any;
+  reproStepsListItemNumberingTitle: any;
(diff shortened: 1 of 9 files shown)


src/native/NativeConstants.ts
```diff
@@ -176,6 +176,7 @@ interface NativeStringKey {
   reproStepsListEmptyStateDescription: any;
   reproStepsListHeader: any;
   reproStepsListItemTitle: any;
+  reproStepsListItemNumberingTitle: any;
```
asamiz, line 179: I have a question just for understanding, why did you give `reproStepsListItemNumberingTitle` the type `any`? Can it be anything else than `string`?

a7medev: Technically, we set it to a string in the native `ArgsRegistry`. So I think it's a valid point to change it to `string` (I'd say we also need to do the same for all the other keys, we should see what they map to and give them an appropriate type).
To not get out of the scope of this PR, I'll only change the type of this key for now, but let's look into changing all the other types as well.

a7medev: @asamiz Looks like TypeScript doesn't like computed string values in enums. Here's the error I get:
```
src/utils/Enums.ts:200:38 - error TS18033: Only numeric enums can have computed members, but this expression has type 'string'. If you do not need exhaustiveness checks, consider using an object literal instead.
```

Looking it up, looks like we either need to upgrade to TypeScript v5 or use an object instead of an enum.

I'll revert it back to `any` but let's look further into it and see what approach we will go with.

asamiz (approved):
Approved, good job @a7medev as usual

PR #1004, opened 2023-07-25
[MOB-12515] Support React Native 0.73

asamiz:
## Description of the change

Add SDK project `namespace` in `build.gradle` file to support the upcoming version of React Native(0.73) as it will use Android Gradle Plugin (AGP) version 8.x which will need the `namespace` to be specified. [More here](https://github.com/react-native-community/discussions-and-proposals/issues/671)

## Type of change

## Related issues

> Issue links go here

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -10,12 +10,15 @@ def shouldUseNameSpace = agpVersion >= 7
 def PACKAGE_PROP = "package=\"com.instabug.reactlibrary\""
 def manifestOutFile = file("${projectDir}/src/main/AndroidManifest.xml")
 def manifestContent = manifestOutFile.getText()
+def isManifestContentUpdated = manifestOutFile.getText() != manifestContent
+def isPackageNamespaceMissing = !manifestContent.contains("$PACKAGE_PROP")
 
 String getExtOrDefault(String name) {
+    def defaultPropertyKey = 'InstabugReactNative_' + name
     if (rootProject.ext.has(name)) {
         return rootProject.ext.get(name)
     }
-    return project.properties['InstabugReactNative_' + name]
+    return project.properties[defaultPropertyKey]
 }
 
 if(shouldUseNameSpace){
@@ -24,7 +27,7 @@ if(shouldUseNameSpace){
         ''
     )  
 } else {
-    if(!manifestContent.contains("$PACKAGE_PROP")){
+    if(isPackageNamespaceMissing){
         manifestContent = manifestContent.replace(
             '<manifest',
             "<manifest $PACKAGE_PROP "
@@ -33,7 +36,10 @@ if(shouldUseNameSpace){
 }
 
 manifestContent.replaceAll("  ", " ")
-manifestOutFile.write(manifestContent)
+
+if(isManifestContentUpdated){
+    manifestOutFile.write(manifestContent)
+}
 
 android {
     if (shouldUseNameSpace){
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -4,6 +4,13 @@ apply from: './jacoco.gradle'
 apply from: './native.gradle'
 apply from: './sourcemaps.gradle'
 
+/* Checking APG version to be backward-compatible with RN versions < 0.71 */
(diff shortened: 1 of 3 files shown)


android/build.gradle
```diff
@@ -4,14 +4,41 @@ apply from: './jacoco.gradle'
...
     return project.properties['InstabugReactNative_' + name]
 }
 
+if(shouldUseNameSpace){
+      manifestContent = manifestContent.replaceAll(
+        PACKAGE_PROP,
+        ''
+    )  
+} else {
+    if(!manifestContent.contains("$PACKAGE_PROP")){
```
abdelhamid-f-nasser, line 27: How about we add an `isPackageNamespaceMissing` instead of `!manifestContent.contains("$PACKAGE_PROP"` property? It would make the logic more transparent and help us understand the condition better.

android/build.gradle
```diff
@@ -4,14 +4,41 @@ apply from: './jacoco.gradle'
...
+        manifestContent = manifestContent.replace(
+            '<manifest',
+            "<manifest $PACKAGE_PROP "
+        )
+    }
+}
+
+manifestContent.replaceAll("  ", " ")
+manifestOutFile.write(manifestContent)
```
abdelhamid-f-nasser, line 36: a suggestion would be to write to manifest only if we made a change to it. We can check for change then write to the manifestFile.

```
if (manifestOutFile.getText() != manifestContent) {
    manifestOutFile.write(manifestContent)
}
```

We can also add a new property `updatedManifestContent` for better readability and use it in checking for changes before writing to the `manifestOutFile`.

asamiz: Yes, I agree with that
Will resolve this and push a new update.

android/build.gradle
```diff
@@ -4,14 +4,41 @@ apply from: './jacoco.gradle'
...
+def shouldUseNameSpace = agpVersion >= 7
+def PACKAGE_PROP = "package=\"com.instabug.reactlibrary\""
+def manifestOutFile = file("${projectDir}/src/main/AndroidManifest.xml")
+def manifestContent = manifestOutFile.getText()
+
 String getExtOrDefault(String name) {
     if (rootProject.ext.has(name)) {
         return rootProject.ext.get(name)
     }
     return project.properties['InstabugReactNative_' + name]
```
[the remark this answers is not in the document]
asamiz: I didn't get it, what do you mean by adding this in a new property?

android/build.gradle
asamiz: Agree with that, I will push new update.

android/build.gradle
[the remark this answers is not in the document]
abdelhamid-f-nasser: like `def defaultValue = 'InstabugReactNative_' + name` so we can substitute it in the 
`project.properties[defaultValue]` or we can also name it `defaultPropertyKey` as I think it's the key to finding our property. I'm a bit confused about this one and what it does so I apologize that it's a bit vague.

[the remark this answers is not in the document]
asamiz: I got you, okay I will push a new update.

abdelhamid-f-nasser (approved):
Outstanding work ! 🎉 you've made the script even cleaner ! 👌🏼

asamiz:
@a7medev Need your review to close the PR.

android/build.gradle
```diff
@@ -4,14 +4,47 @@ apply from: './jacoco.gradle'
...
+        ''
+    )  
+} else {
+    if(isPackageNamespaceMissing){
+        manifestContent = manifestContent.replace(
+            '<manifest',
+            "<manifest $PACKAGE_PROP "
+        )
+    }
+}
```
a7medev, line 36: @asamiz I think we can change this to `else if` instead, what do you think?

a7medev (changes requested):
Great job @asamiz! and thanks @abdelhamid-f-nasser for the detailed review 🙏🏼 I just have small one comment that needs to get resolved 🙏🏼

a7medev (approved):
Great job @asamiz! 🚀❤️

PR #1012, opened 2023-08-22
feat: cp rn report startup crashes on android

abdelhamid-f-nasser:
## Description of the change

- Add native-side init API which can be used to catch and report startup crashes on android.
- Refactor  the react-native init to reuse the native-side#init method
- Cover code with tests

## Type of change

## Related issues

> [MOB-12628](https://instabug.atlassian.net/browse/MOB-12628), [MOB-12689](https://instabug.atlassian.net/browse/MOB-12689), [MOB-12690](https://instabug.atlassian.net/browse/MOB-12690)

### Development

- [ ] Lint rules pass locally

### Code review


android/src/test/java/com/instabug/reactlibrary/RNInstabugReactnativeModuleTest.java
```diff
@@ -452,7 +452,7 @@ public void tearDown() {
     @Test
     public void givenString$setString_whenQuery_thenShouldCallNativeApiWithEnum() {
         // given
-        mockStatic(Log.class);
+//        mockStatic(Log.class);
```
a7medev, line 455: Let's remove this line instead of commenting it.

android/build.gradle
```diff
@@ -46,6 +46,11 @@ android {
     compileSdkVersion getExtOrDefault('compileSdkVersion').toInteger()
     buildToolsVersion getExtOrDefault('buildToolsVersion')
 
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
+
```
a7medev, line 53: I don't think we should upgrade to Java 8 at the moment.
Let's revert back to the original version we were using (I think this means we will also need to replace lambdas with anonymous classes and probably other changes).
I definitely think we should take this decision of just using Java 8 and migrate to new features like lambda expressions some day but we shouldn't do that on the fly as part of another PR without carefully evaluating it.

android/src/main/java/com/instabug/reactlibrary/RNInstabug.java
```diff
@@ -0,0 +1,135 @@
...
+     *
+     * <blockquote><pre>
+     * RNInstabug.initNativeSdk(
+     *     this,
+     *     "f23c....",
+     *     new InstabugInvocationEvent[]{
+     *         InstabugInvocationEvent.NONE,
+     *     },
+     *     ""
+     * );
```
a7medev, line 78: Let's update these examples to match the new signature.

android/src/main/java/com/instabug/reactlibrary/RNInstabug.java
```diff
@@ -0,0 +1,135 @@
...
+     * </pre></blockquote>
+     *
+     */
+    public void init(
+            @NonNull Application application,
+            @NonNull String applicationToken,
+            int logLevel,
+            @NonNull InstabugInvocationEvent... instabugInvocationEvent
```
a7medev, line 86: I think the log level should not be required, let's change the method signature to make it optional. What do you think?

android/src/main/java/com/instabug/reactlibrary/RNInstabug.java
```diff
@@ -0,0 +1,135 @@
...
+
+    @VisibleForTesting
+    public void setCurrentPlatform() {
+        try {
+            Method method = InstabugUtil.getMethod(Class.forName("com.instabug.library.Instabug"), "setCurrentPlatform", int.class);
+            if (method != null) {
+                Log.i("IB-CP-Bridge", "invoking setCurrentPlatform with platform: " + Platform.RN);
+                method.invoke(null, Platform.RN);
+            } else {
+                Log.e("IB-CP-Bridge", "setCurrentPlatform was not found by reflection");
```
a7medev, line 113: We usually use `IBG` so let's match that here as well.

```suggestion
                Log.i("IBG-CP-Bridge", "invoking setCurrentPlatform with platform: " + Platform.RN);
                method.invoke(null, Platform.RN);
            } else {
                Log.e("IBG-CP-Bridge", "setCurrentPlatform was not found by reflection");
```

android/src/main/java/com/instabug/reactlibrary/RNInstabug.java
```diff
@@ -0,0 +1,135 @@
...
+            if (method != null) {
+                method.invoke(null, "https://docs.instabug.com/docs/react-native-sdk-migration-guide");
+            }
+        } catch (ClassNotFoundException e) {
+            e.printStackTrace();
+        } catch (IllegalAccessException e) {
+            e.printStackTrace();
+        } catch (InvocationTargetException e) {
+            e.printStackTrace();
+        }
```
a7medev, line 133: Let's group all these in a single `catch` block, what do you think?

a7medev (changes requested):
Great effort @abdelhamid-f-nasser! 👏🏼 Please check the review comments. 🙏🏼

android/src/main/java/com/instabug/reactlibrary/RNInstabug.java
abdelhamid-f-nasser: As we discussed we've introduced an init method without the `logLevel` param and it passes `LogLevel.ERROR` as default.

PR #1015, opened 2023-08-30
ci: automate nn releases

asamiz:
## Description of the change

Automate SDK changes for NN enterprise release.

## Type of change

## Related issues

[MOB-12644](https://instabug.atlassian.net/browse/MOB-12644)

### Development

- [ ] Lint rules pass locally

### Code review

[MOB-12644]: https://instabug.atlassian.net/browse/MOB-12644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -330,9 +330,9 @@ workflows:
             - e2e_ios
             - e2e_android
           type: approval
-          # filters:
-          #   branches:
-          #     only: master
+          filters:
+            branches:
+              only: master
       - publish:
           requires:
             - hold_publish
@@ -342,6 +342,6 @@ workflows:
       - release_nn:
           requires:
             - hold_release_nn
-          # filters:
-          #   branches:
-          #     only: master
+          filters:
+            branches:
+              only: master


abdelhamid-f-nasser:
Great effort👌🏼, I love how you documented the script to make it more readable.

Just one tip, you can better change the PR title to match the conventions for example:
<type>(scope): commit title (all lowercase)
- `chore: automate nn releases`
- `build: automate nn releases`

*Note: the scopes we currently work with are `android` for android related changes and `iOS` otherwise we just omit it.
reference: [Git guidelines](https://instabug.atlassian.net/wiki/spaces/CP/pages/3700129799/Git+GitHub+Guidelines)*

abdelhamid-f-nasser (approved):
Nicely done @asamiz as always !! 💪 🚀

.circleci/config.yml
```diff
@@ -224,6 +235,36 @@ jobs:
           working_directory: examples/default
           command: detox test -c android.emu.release
 
+  # Automate the enterprise NN sdk changes
+  release_nn:
+    working_directory: ~/project
+    macos:
+      xcode: 13.4.1
```
a7medev, line 242: Do we need to run it on a macOS machine? If not, I think we can switch to a `node/default` executor instead since it's usually faster and cheaper.
I believe switching to `node/default` which runs on Linux would also break the current `sed` command and we might need to remove the `"" -e` part of the command so let's make sure everything is still working as expected.

a7medev (changes requested):
Great job @asamiz! 👏🏼 Just some small changes and we're ready to go! 🚀

.circleci/config.yml
a7medev: AFAIK the `sed` command is a bit different on macOS as you need to add `""` which isn't needed on Linux, we just need to make sure that we update the job .

a7medev (approved):

PR #1016, opened 2023-09-03
refactor: use anonymous class instead of lambda

a7medev:
## Description of the change

Removed the use of Java 8 lambda functions in Android that was added in #1012 and replaced it with `new Runnable` since it breaks apps running on older versions of Java and React Native.

## Type of change

## Related issues

Jira ID: MOB-12977

### Development

- [ ] Lint rules pass locally

### Code review


abdelhamid-f-nasser (approved):
Thanks for covering this 🙏🏼. I'll make sure to use anon classes as closure breaks with old java versions.

Reference: We can't have lambdas anywhere in our Java codebase (except in tests) since it'd break apps on older versions of Java.

PR #1044, opened 2023-10-25
fix: fix and enhance AGP namespace support

a7medev:
## Description of the change

The current implementation of Android Gradle Plugin >= 7.3.0 namespace support calculates `isManifestContentUpdated` before applying any modifications to the file, so the manifest file was never updated since `isManifestContentUpdated` will always be `false`. The fix was to calculate it right before writing the updates to the file.

I've changed the namespace support condition to check for AGP version >= 7.3 not only >= 7 since `namespace` support was added in v7.3.0. I've also refactored the code into functions and renamed some variables for readability.

I've also kept the `package` property in the initial `AndroidManifest.xml` file since having it removed by default causes the first app build to fail sometimes with apps not using AGP >= 7.3.0, while having it there in the initial run when it's not needed (when using `namespace`) won't cause trouble.

I've tested these changes on React Native `0.66.0` and `0.73.0-rc.3` and its working fine.

## Type of change

## Related issues

Jira ID: INSD-10287

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased](https://github.com/Instabug/Instabug-React-Native/compare/v12.1.0...dev)
+
+### Fixed
+
+- Fix an issue with Android Gradle Plugin namespace support required for React Native 0.73 and backward compatibility with previous versions ([#1044](https://github.com/Instabug/Instabug-React-Native/pull/1044)).
+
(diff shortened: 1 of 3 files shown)


abdelhamid-f-nasser (approved):
The code looks excellent and is very self-explanatory 💪 !! Great effort supporting it @a7medev 👍

PR #1166, opened 2024-03-13
fix: resolve syntax error in old TypeScript versions

### LINKED ISSUE #1161: Error loading JS bundle when running on Android
A build error is introduced since 12.7.0, please see screenshot below.

Screenshot_1709550396

Did you test the library before you publish the new version?

### SDK Version

"instabug-reactnative": "12.7.0",

### Device Model

Android simulator

### [Optional] Project That Reproduces the Issue

just upgrade instabug-reactnative to 12.7.0, build on Android simulator successfully and then failed to load the js bundle because of this error 

```
error node_modules/instabug-reactnative/src/modules/Instabug.ts: Unexpected token: operator (?) in file node_modules/instabug-reactnative/src/modules/Instabug.ts at 52:36. Run CLI with --verbose flag for more details.
```

This is the screenshot of the code in `instabug-reactnative`

Screenshot 2024-03-04 at 19 09 42

ahmedAlaaInstabug:
## Description of the change

1. fix unExpected token (??) on old node version

## Type of change

## Related issues

Fixes #1161 
Jira ID = INSD-11129

### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 - Add `SessionReplay.getSessionReplayLink` API which retrieves the current session's replay link ([#1142](https://github.com/Instabug/Instabug-React-Native/pull/1142)).
 - Support setting the Code Push version after SDK initialization ([#1143](https://github.com/Instabug/Instabug-React-Native/pull/1143)).
 
+### Changed
+
+- Bump Instabug Android SDK to v12.8.0 ([#1149](https://github.com/Instabug/Instabug-React-Native/pull/1149)). [See release notes](https://github.com/Instabug/android/releases/tag/v12.8.0).
+
 ## [12.7.1](https://github.com/Instabug/Instabug-React-Native/compare/v12.7.0...v12.7.1) (February 15, 2024)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,6 @@
 
 - Fix resolve syntax error (??) in old TypeScript versions ([#1166](https://github.com/Instabug/Instabug-React-Native/pull/1166)).
 
-
 ## [12.8.0](https://github.com/Instabug/Instabug-React-Native/compare/v12.7.1...v12.8.0) (February 25, 2024)
 
 ### Added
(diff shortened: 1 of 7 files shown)


a7medev (changes requested):
LGTM 👌🏼. Let's add a changelog item and we'd be good to go.

Let's also link it to a Jira card.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## [Unreleased](https://github.com/Instabug/Instabug-React-Native/compare/v12.8.0...dev)
+
+### Fixed
+
+- Fix resolve syntax error (??) in old TypeScript versions ([#1166](https://github.com/Instabug/Instabug-React-Native/pull/1166)).
```
a7medev, line 7: I'd recommend something like this, what do you think?

```suggestion
- Remove the use of the nullish coalescing assignment operator (`??=`) causing a syntax error with older TypeScript versions ([#1166](https://github.com/Instabug/Instabug-React-Native/pull/1166)), closes [#1161
](https://github.com/Instabug/Instabug-React-Native/issues/1161).
```

a7medev (changes requested):

src/modules/Instabug.ts
```diff
@@ -64,7 +64,9 @@ export const init = (config: InstabugConfig) => {
   captureUnhandledRejections();
 
   // Default networkInterceptionMode to JavaScript
-  config.networkInterceptionMode ??= NetworkInterceptionMode.javascript;
+  if (!config.networkInterceptionMode) {
```
a7medev, line 67: Let's change this if condition to explicitly check for `null` or `undefined` instead of `!config.networkInterceptionMode` since the enum values will get mapped to integer values with the first enum value being assigned the number `0` so if a user sets the `networkInterceptionMode` to the first enum value JavaScript will treat it as `false` thus changing the user's intended value to the default one even when it's actually set in the user's code.

In this case the first enum value is `javascript` which is the current default so it won't cause bugs, but if we changed the default to be something else users won't be able to set the interception mode to the first enum value as explained above.

```suggestion
  // The `== null` here acts as `=== null or === undefined`
  if (config.networkInterceptionMode == null) {
```

a7medev (changes requested):

PR #1194, opened 2024-04-25
feat: enhance non-fatals support 

ahmedAlaaInstabug:
## Description of the change

1. add non fatal api

## Type of change

## Related issues

JIRA ID :

### Development

- [ ] Lint rules pass locally

### Code review


src/modules/CrashReporting.ts
```diff
@@ -16,9 +17,23 @@ export const setEnabled = (isEnabled: boolean) => {
...
+ * @param levelType different severity levels for errors
  */
-export const reportError = (error: ExtendedError) => {
-  return InstabugUtils.sendCrashReport(error, NativeCrashReporting.sendHandledJSCrash);
+export const reportError = (
+  error: ExtendedError,
+  userAttributes: Object | null,
+  fingerprint: string | null,
+  levelType: NonFatalErrorType,
+) => {
```
a7medev, line 29: We need to make `userAttributes`, `fingerprint`, `levelType` (I think we can rename it to `level`) optional, we should also turn them into an `options` object or something similar to named parameters in other languages for clarity as the function now has 4 parameters. Let's also turn `userAttributes` into `Record<string, string>` since `Object` would match too many unwanted types.

We will need to define an interface `NonFatalOptions` for the options, we can put it in `models` as we do with `InstabugConfig` and export it from `src/index.ts` to be accessible to users.

Here's how it might look.

```ts
interface NonFatalOptions {
  userAttributes?: Record<string, string>;
  fingerprint?: string;
  level?: NonFatalErrorLevel;
}
```

```suggestion
export const reportError = (
  error: ExtendedError,
  { userAttributes, fingerprint, level }: NonFatalOptions = {},
) => {
```

android/src/main/java/com/instabug/reactlibrary/RNInstabugCrashReportingModule.java
```diff
@@ -79,41 +85,61 @@ public void run() {
      * Send handled JS error object
      *
      * @param exceptionObject Exception object to be sent to Instabug's servers
+     * @param userAttributes  (Optional) extra user attributes attached to the crash
+     * @param fingerprint     (Optional) key used to customize how crashes are grouped together
+     * @param levelType       different severity levels for errors
      */
     @ReactMethod
-    public void sendHandledJSCrash(final String exceptionObject) {
+    public void sendHandledJSCrash(final String exceptionObject, @androidx.annotation.Nullable ReadableMap userAttributes, @androidx.annotation.Nullable String fingerprint, String levelType) {
```
a7medev, line 93: Let's import `Nullable` instead of `androidx.annotation.Nullable` for readability.

We can also turn `level` into `Nullable` as well since we already default to level of error.

```suggestion
    public void sendHandledJSCrash(final String exceptionObject, @Nullable ReadableMap userAttributes, @Nullable String fingerprint, @Nullable String level) {
```

examples/default/src/screens/CrashReportingScreen.tsx
```diff
@@ -46,6 +50,48 @@ export const CrashReportingScreen: React.FC = () => {
...
+    inputWrapper: {
+      padding: 4,
+      flex: 1,
+    },
+
+    formContainer: {
+      flexDirection: 'row',
+      alignItems: 'stretch',
+    },
+  });
```
a7medev, line 63: We should move these `styles` outside the component function to not get recreated with each render. We usually define it under the component function.

examples/default/src/screens/CrashReportingScreen.tsx
```diff
@@ -46,6 +50,48 @@ export const CrashReportingScreen: React.FC = () => {
...
+  const [userAttributeValue, setUserAttributeValue] = useState('');
+  const [crashNameValue, setCrashNameValue] = useState('');
+  const [crashFingerprint, setCrashFingerprint] = useState('');
+  const [crashLevelValue, setCrashLevelValue] = useState<NonFatalErrorType>(
+    NonFatalErrorType.error,
+  );
+
+  function sendCrash() {
+    try {
+      const error = Error(crashNameValue);
```
a7medev, line 74: This should be `new Error` I think.

```suggestion
      const error = new Error(crashNameValue);
```

examples/default/src/screens/CrashReportingScreen.tsx
```diff
@@ -46,6 +50,48 @@ export const CrashReportingScreen: React.FC = () => {
...
+      if (!error.message) {
+        const appName = 'Instabug Test App';
+        error.message = `Handled ${crashNameValue} From ${appName}`;
+      }
+      throw error;
+    } catch (err) {
```
a7medev, line 80: We are already passing `crashNameValue` to `new Error` as a message (the first argument to the `Error` constructor is the message), so `error.message` will be empty only when `crashNameValue` is empty. I think we can just put the message we want to put in the constructor, no need for this check in my opinion.

examples/default/src/screens/CrashReportingScreen.tsx
```diff
@@ -46,6 +50,48 @@ export const CrashReportingScreen: React.FC = () => {
...
+      if (err instanceof Error) {
+        const attrMap: { [k: string]: string } = {};
+        attrMap[userAttributeKey] = userAttributeValue;
+
+        CrashReporting.reportError(
+          err,
+          userAttributeValue.length === 0 || userAttributeKey.length === 0 ? null : attrMap,
+          crashFingerprint.length === 0 ? null : crashFingerprint,
+          crashLevelValue,
+        ).then(() => Alert.alert(`Crash report for ${crashNameValue} is Sent!`));
```
a7medev, line 90: I think it'd be more readable to use `Record<string, string>` for `userAttributes` and simplify the conditions a bit.

I feel like the API signature is a bit verbose, so hopefully it gets better after the proposed change.

```suggestion
        const userAttributes: Record<string, string> = {};
        if (userAttributesKey && userAttributesValue) {
          userAttributes[userAttributesKey] = userAttributesValue;
        }

        CrashReporting.reportError(
          err,
          userAttributes,
          crashFingerprint || undefined,
          crashLevelValue,
        ).then(() => Alert.alert(`Crash report for ${crashNameValue} is Sent!`));
```

test/modules/CrashReporting.spec.ts
```diff
@@ -15,11 +16,17 @@ describe('CrashReporting Module', () => {
 
   it('should call the native method sendCrashReporting with JSON object and sendHandledJsCrash as a callback', () => {
     const error = new TypeError('Invalid type');
-    CrashReporting.reportError(error);
+    const userAttribute: Map<string, string> | null = null;
```
a7medev, line 19: `Map<string, string>` is different from object literals (and `Record<string, string>` which is an object literal with a `string` key and `string` value), so the type of `userAttributes` here doesn't match what it is intended to be.

We can't pass [JavaScript Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)s to native methods, but we can pass object literals, so if we actually use `Map` in a real app it'd cause a crash.

a7medev (changes requested):

test/modules/CrashReporting.spec.ts
```diff
@@ -15,11 +16,17 @@ describe('CrashReporting Module', () => {
 
   it('should call the native method sendCrashReporting with JSON object and sendHandledJsCrash as a callback', () => {
```
ahmedAlaaInstabug, line 17: j

android/src/main/java/com/instabug/reactlibrary/RNInstabugCrashReportingModule.java
```diff
@@ -79,41 +85,61 @@ public void run() {
...
-        } catch (Exception e) {
+            MainThreadHandler.runOnMainThread(new Runnable() {
+                @Override
+                public void run() {
+                    try {
+                        Method method = getMethod(Class.forName("com.instabug.crash.CrashReporting"), "reportException", JSONObject.class, boolean.class,
+                                Map.class, JSONObject.class, IBGNonFatalException.Level.class);
+                        if (method != null) {
+                            IBGNonFatalException.Level nonFatalExceptionLevelType = ArgsRegistry.nonFatalExceptionLevel.getOrDefault(levelType, IBGNonFatalException.Level.ERROR);
+                            method.invoke(null, jsonObject, true, userAttributes == null ? null : userAttributes.toHashMap(), fingerprint == null ? null : CrashReporting.getFingerprintObject(fingerprint), nonFatalExceptionLevelType);
```
a7medev, line 104: I feel like it's a bit cluttered to have all these ternary statements in the arguments, why don't we extract them into variables instead and then just pass the variables to `method.invoke`?

a7medev (changes requested):
Good job @ahmedAlaaInstabug! 👏🏼 Just a couple of little comments and we should be good to go!

examples/default/src/screens/CrashReportingScreen.tsx
```diff
@@ -50,44 +62,35 @@ export const CrashReportingScreen: React.FC = () => {
...
-        const appName = 'Instabug Test App';
-        error.message = `Handled ${crashNameValue} From ${appName}`;
-      }
+      const error = new Error(crashNameValue);
+
       throw error;
     } catch (err) {
       if (err instanceof Error) {
         const attrMap: { [k: string]: string } = {};
         attrMap[userAttributeKey] = userAttributeValue;
```
a7medev: Let's remove the `attrMap` variable since it's now unused.

a7medev (dismissed):
Perfect! Let's gooo! 🚀🚀

PR #1243, opened 2024-06-30
fix: w3c ios header injection

kholood-ea:
## Description of the change
fix IOS W3C header injection

## Type of change

## Related issues

[jiraID:MOB-14775](https://instabug.atlassian.net/browse/MOB-14775)
### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugAPMModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugAPMModule.java
@@ -408,7 +408,7 @@ public class RNInstabugAPMModule extends EventEmitterModule {
                         public void invoke(@NonNull APMFeaturesAvailability apmFeaturesAvailability) {
                             WritableMap params = Arguments.createMap();
                             params.putBoolean("isW3ExternalTraceIDEnabled", apmFeaturesAvailability.isW3CExternalTraceIdAvailable());
-                            params.putBoolean("isW3ExternalGeneratedHeaderEnabled", apmFeaturesAvailability.getShouldAttachGeneratedHeader()); // Example boolean value
+                            params.putBoolean("isW3ExternalGeneratedHeaderEnabled", apmFeaturesAvailability.getShouldAttachGeneratedHeader());
                             params.putBoolean("isW3CaughtHeaderEnabled", apmFeaturesAvailability.getShouldAttachCapturedHeader());
 
                             sendEvent(Constants.IBGAPM_ON_NEW_W3C_FLAGS_UPDATE_RECEIVED_CALLBACK, params);
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugAPMModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugAPMModule.java
@@ -5,27 +5,27 @@ import android.os.SystemClock;
 import android.util.Log;
 
 import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
 
+import com.facebook.react.bridge.Arguments;
+import com.facebook.react.bridge.Callback;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
-import com.facebook.react.bridge.ReactContext;
-import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.WritableMap;
 import com.instabug.apm.APM;
 import com.instabug.apm.configuration.cp.APMFeature;
+import com.instabug.apm.configuration.cp.FeatureAvailabilityCallback;
 import com.instabug.apm.model.ExecutionTrace;
 import com.instabug.apm.networking.APMNetworkLogger;
 import com.instabug.apm.networkinterception.cp.APMCPNetworkLog;
 import com.instabug.apm.configuration.cp.FeaturesChangeListener;
 import com.instabug.apm.configuration.cp.APMFeaturesAvailability;
-import com.instabug.chat.Replies;
+import com.instabug.reactlibrary.utils.EventEmitterModule;
 import com.instabug.reactlibrary.utils.MainThreadHandler;
 import com.instabug.apm.InternalAPM;
 
 import org.json.JSONException;
 import org.json.JSONObject;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 
 import java.util.HashMap;
@@ -34,7 +34,7 @@ import javax.annotation.Nonnull;
 
 import static com.instabug.reactlibrary.utils.InstabugUtil.getMethod;
 
-public class RNInstabugAPMModule extends ReactContextBaseJavaModule {
+public class RNInstabugAPMModule extends EventEmitterModule {
 
     public RNInstabugAPMModule(ReactApplicationContext reactApplicationContext) {
         super(reactApplicationContext);
@@ -47,6 +47,16 @@ public class RNInstabugAPMModule extends ReactContextBaseJavaModule {
     @Override
     public String getName() {
         return "IBGAPM";
+    }
+
+    @ReactMethod
+    public void addListener(String event) {
+        super.addListener(event);
+    }
+
+    @ReactMethod
+    public void removeListeners(Integer count) {
+        super.removeListeners(count);
     }
 
     /**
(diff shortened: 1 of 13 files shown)


test/utils/XhrNetworkInterceptor.spec.ts
```diff
@@ -372,13 +375,14 @@ describe('Network Interceptor W3C Headers', () => {
...
-      expect(network.wceti).toBe(null);
-      expect(network.requestHeaders).not.toHaveProperty('traceparent');
+      injectHeaders(network, featureFlags).then((res) => {
+        expect(res).toBeUndefined();
+        expect(network.w3cc).toBe(null);
+        expect(network.partialId).toBe(null);
+        expect(network.etst).toBe(null);
+        expect(network.wgeti).toBe(null);
+        expect(network.wceti).toBe(null);
+      });
```
a7medev, line 385: I think using async/await would be more readable here and in the other similar tests, what do you think?

a7medev (changes requested):

android/src/main/java/com/instabug/reactlibrary/RNInstabugAPMModule.java
```diff
@@ -335,9 +335,30 @@ public void run() {
...
     */
     @ReactMethod
-    public void networkLog(String networkData) throws JSONException {
+    public void networkLog(String networkData, Boolean isW3cHeaderFound,Double partialId,Double networkStartTimeInSeconds,String W3CGeneratedHeader,String W3CCaughtHeader) throws JSONException {
         try{
             APMNetworkLogger apmNetworkLogger = new APMNetworkLogger();
+
+            Long NetworkStartTime = null;
+            Long PartialId=null;
+            String GeneratedHeader=null;
```
a7medev, line 344: Let's rename all the variables to use `camelCase` instead of `PascalCase`, for ones starting with `W3C` like `W3CGeneratedHeader` I think we can rename them to start with `w3c` like `w3cGeneratedHeader`.

src/utils/XhrNetworkInterceptor.ts
```diff
@@ -24,11 +24,11 @@ export interface NetworkData {
...
   serverErrorMessage: string;
   requestContentType: string;
-  w3cc: boolean | null;
-  partialId: number | null;
-  etst: number | null;
-  wgeti: string | null;
-  wceti: string | null;
+  isW3cHeaderFound: boolean;
+  partialId: number;
+  networkStartTimeInSeconds: number;
```
a7medev, line 29: We already collect the start time in `startTime` but in milliseconds, so I think we can just use `network.startTime / 1000` to get the time in seconds instead of storing it in a different field to avoid discrepancy.

src/utils/XhrNetworkInterceptor.ts
```diff
@@ -24,11 +24,11 @@ export interface NetworkData {
...
+  W3CgeneratedHeader: string;
+  W3CCaughtHeader: string;
```
a7medev, line 31: Let's use `camelCase` for these.
```suggestion
  w3cGeneratedHeader: string;
  w3cCaughtHeader: string;
```

src/utils/XhrNetworkInterceptor.ts
```diff
@@ -98,26 +99,33 @@ export const injectHeaders = async (
...
+  networkData.isW3cHeaderFound = headerFound;
+  const injectionMethodology = headerFound
     ? IdentifyCaughtHeader(networkData, w3c_caught_header)
     : injectGeneratedData(networkData, w3c_generated_header);
+  return injectionMethodology;
 };
 
-const IdentifyCaughtHeader = (networkData: NetworkData, w3c_caught_header: boolean) => {
-  w3c_caught_header && (networkData.wceti = networkData.requestHeaders.traceparent);
+const IdentifyCaughtHeader = async (networkData: NetworkData, w3c_caught_header: boolean) => {
```
a7medev, line 109: Let's follow `camelCase` here as well.
```suggestion
const identifyCaughtHeader = async (networkData: NetworkData, w3cCaughtHeader: boolean) => {
```

src/utils/XhrNetworkInterceptor.ts
```diff
@@ -98,26 +99,33 @@ export const injectHeaders = async (
...
+  if (w3c_caught_header) {
+    networkData.W3CCaughtHeader = networkData.requestHeaders.traceparent;
+    return networkData.requestHeaders.traceparent;
+  } else {
+    return;
+  }
```
a7medev, line 115: The `else` block here is not needed, we can just remove it.

src/utils/XhrNetworkInterceptor.ts
```diff
@@ -98,26 +99,33 @@ export const injectHeaders = async (
...
+  networkData.networkStartTimeInSeconds = timestampInSeconds;
 
   if (w3c_generated_header) {
-    networkData.wgeti = w3cHeader;
-    networkData.requestHeaders.traceparent = w3cHeader;
+    networkData.W3CgeneratedHeader = w3cHeader;
+    return w3cHeader;
+  } else {
+    return;
   }
```
a7medev, line 128: Same for the `else` block here as well, we can just remove it as there's no logic after it.

ios/RNInstabug/InstabugReactBridge.m
```diff
@@ -283,7 +283,15 @@ - (dispatch_queue_t)methodQueue {
     }
 }
 
-RCT_EXPORT_METHOD(networkLog:(NSDictionary *) networkData) {
+RCT_EXPORT_METHOD(networkLog:(NSDictionary *) networkData isW3cCaughted:(nonnull NSNumber  *)isW3cCaughted partialID:(nonnull NSNumber  *) partialID timestamp:(nonnull NSNumber  *) timestamp generatedW3CTraceparent:(nullable NSString  *) generatedW3CTraceparent caughtedW3CTraceparent:(nullable NSString  *) caughtedW3CTraceparent){
```
a7medev, line 286: I think the parameters are getting really long here so we can move each parameter to its own line.
```suggestion
RCT_EXPORT_METHOD(networkLog:(NSDictionary *)networkData
               isW3cCaughted:(nonnull NSNumber  *)isW3cCaughted
                   partialID:(nonnull NSNumber  *)partialID
                   timestamp:(nonnull NSNumber  *)timestamp
     generatedW3CTraceparent:(nullable NSString  *)generatedW3CTraceparent
      caughtedW3CTraceparent:(nullable NSString  *)caughtedW3CTraceparent) {
```

a7medev, line 286: Also there's a typo in `caughtedW3CTraceparent`, should be `caughtW3CTraceparent`

ios/RNInstabug/InstabugReactBridge.m
```diff
@@ -283,7 +283,15 @@ - (dispatch_queue_t)methodQueue {
...
+    
+    if ([partialID isEqualToNumber:@0]) {
+        partialID=nil;
+        timestamp=nil;
+    }
+    if([generatedW3CTraceparent isEqualToString:@""]){generatedW3CTraceparent=nil;}
+    if([caughtedW3CTraceparent isEqualToString:@""]){caughtedW3CTraceparent=nil;}
```
a7medev, line 293: Let's please format these two lines.

a7medev (changes requested):

src/utils/XhrNetworkInterceptor.ts
kholood-ea: network start time in seconds are used in many places so its better to call networkStartTimeInSeconds rather than do a calculation every time

a7medev: I feel like a getter would be more suitable here instead of a different attribute since now the values could easily go out of sync.

We could do something like this:

```typescript
interface NetworkData {
  // ...
  startTime: number;
  startTimeInSeconds: number;
}

const network: Network = {
  // ...
  startTime: 12000,
  get startTimeInSeconds() {
    return Math.floor(this.startTime / 1000);
  },
};
```

src/utils/XhrNetworkInterceptor.ts
```diff
@@ -271,7 +277,8 @@ export default {
       }
 
       cloneNetwork.startTime = Date.now();
-      getFeatureFlags(cloneNetwork);
+      const traceparent = await getFeatureFlags(cloneNetwork);
+      traceparent && this.setRequestHeader('Traceparent', traceparent);
```
a7medev, line 281:
```suggestion
      if (traceparent) {
        this.setRequestHeader('traceparent', traceparent);
      }
```

src/utils/XhrNetworkInterceptor.ts
```diff
@@ -98,26 +99,31 @@ export const injectHeaders = async (
...
+  networkData.isW3cHeaderFound = headerFound;
+  const injectionMethodology = headerFound
+    ? identifyCaughtHeader(networkData, w3c_caught_header)
     : injectGeneratedData(networkData, w3c_generated_header);
+  return injectionMethodology;
 };
 
-const IdentifyCaughtHeader = (networkData: NetworkData, w3c_caught_header: boolean) => {
-  w3c_caught_header && (networkData.wceti = networkData.requestHeaders.traceparent);
+const identifyCaughtHeader = async (networkData: NetworkData, w3c_caught_header: boolean) => {
```
a7medev, line 109:
```suggestion
const identifyCaughtHeader = async (networkData: NetworkData, isW3cCaughtHeaderEnabled: boolean) => {
```

src/utils/XhrNetworkInterceptor.ts
```diff
@@ -98,26 +99,31 @@ export const injectHeaders = async (
...
+  if (w3c_caught_header) {
+    networkData.w3cCaughtHeader = networkData.requestHeaders.traceparent;
+    return networkData.requestHeaders.traceparent;
+  }
+  return;
 };
 
-const injectGeneratedData = (networkData: NetworkData, w3c_generated_header: boolean) => {
+const injectGeneratedData = async (networkData: NetworkData, w3c_generated_header: boolean) => {
```
a7medev, line 117:
```suggestion
const injectGeneratedData = async (networkData: NetworkData, isW3cExternalGeneratedHeaderEnabled: boolean) => {
```

a7medev (changes requested):

a7medev (approved):
LGTM, let's goo! 🚀

PR #1258, opened 2024-07-25
refactor: move feature flags from APM to core

ahmedAlaaInstabug:
## Description of the change

1) migrate featureflags Apm method to Core

## Type of change

## Related issues

JIRA ID : MOB-14801

### Development

- [ ] Lint rules pass locally

### Code review


android/src/main/java/com/instabug/reactlibrary/Constants.java
```diff
@@ -10,6 +10,6 @@ final class Constants {
     final static String IBG_ON_NEW_MESSAGE_HANDLER = "IBGonNewMessageHandler";
     final static String IBG_ON_NEW_REPLY_RECEIVED_CALLBACK = "IBGOnNewReplyReceivedCallback";
 
-    final static String IBGAPM_ON_NEW_W3C_FLAGS_UPDATE_RECEIVED_CALLBACK = "IBGAPMOnNewW3CFlagsUpdateReceivedCallback";
+    final static String IBGCore_ON_NEW_W3C_FLAGS_UPDATE_RECEIVED_CALLBACK = "IBGCoreOnNewW3CFlagsUpdateReceivedCallback";
```
a7medev, line 13: I think let's rename this to be prefixed with `IBG` instead of `IBGCore` to align with the rest of the constants, what do you think?

```suggestion
    final static String IBG_ON_NEW_W3C_FLAGS_UPDATE_RECEIVED_CALLBACK = "IBGOnNewW3CFlagsUpdateReceivedCallback";
```

src/native/NativeInstabug.ts
```diff
@@ -136,12 +136,29 @@ export interface InstabugNativeModule extends NativeModule {
...
+      isW3CaughtHeaderEnabled: boolean;
+    }) => void,
+  ): void;
 }
 
 export const NativeInstabug = NativeModules.Instabug;
 
 export enum NativeEvents {
   PRESENDING_HANDLER = 'IBGpreSendingHandler',
+  ON_W3C_FLAGS_CHANE = 'IBGCoreOnNewW3CFlagsUpdateReceivedCallback',
```
a7medev, line 161: Let's rename this one as well if you think we should rename the one in the native.

```suggestion
  ON_W3C_FLAGS_CHANE = 'IBGOnNewW3CFlagsUpdateReceivedCallback',
```

a7medev (changes requested):
Good job @ahmedAlaaInstabug 👏🏼
I just have a slight comment so let me know your thoughts 🙏🏼

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
```diff
@@ -1082,6 +1089,105 @@ public void run() {
             }
         });
     }
+    /**
+     * Register a listener for W3C flags value change
+     */
+    @ReactMethod
+    public void registerW3CFlagsChangeListener(final Callback handler){
```
a7medev, line 1096: We are not using this `handler` callback anywhere in the code so we don't need to send it to the native side, I think we can safely remove it.

a7medev (changes requested):

src/modules/Instabug.ts
```diff
@@ -580,3 +580,20 @@ export const componentDidAppearListener = (event: ComponentDidAppearEvent) => {
     _lastScreen = event.componentName;
   }
 };
+
+/**
+ * Sets listener to W3ExternalTraceID flag changes
+ * @param handler A callback that gets the update value of the flag
+ */
+export const _registerW3CFlagsChangeListener = (
```
a7medev, line 588: Let's annotate this with `@internal`. We can also add the `stripInternal` TSConfig flag as `true` to strip `@internal` functions like this from the type definitions, what do you think?

```suggestion
/**
 * Sets listener to W3ExternalTraceID flag changes
 * @param handler A callback that gets the update value of the flag
 * @internal
 */
export const _registerW3CFlagsChangeListener = (
```

PR #1270, opened 2024-08-11
feat: navigation tracking support with expo router

kholood-ea:
## Description of the change
Support navigation tracking with expo router 

## Type of change

## Related issues

[ID:MOB-15377](https://instabug.atlassian.net/browse/MOB-15377)
### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/examples/default/src/App.tsx
+++ b/examples/default/src/App.tsx
@@ -2,7 +2,7 @@ import React, { useEffect } from 'react';
 import { StyleSheet } from 'react-native';
 
 import { GestureHandlerRootView } from 'react-native-gesture-handler';
-import { NavigationContainer } from '@react-navigation/native';
+import { NavigationContainer, useNavigationContainerRef } from '@react-navigation/native';
 import Instabug, {
   CrashReporting,
   InvocationEvent,
@@ -20,6 +20,7 @@ import { QueryClient, QueryClientProvider } from 'react-query';
 const queryClient = new QueryClient();
 
 export const App: React.FC = () => {
+  const navigationRef = useNavigationContainerRef();
   useEffect(() => {
     Instabug.init({
       token: 'deb1910a7342814af4e4c9210c786f35',
@@ -33,11 +34,15 @@ export const App: React.FC = () => {
     });
   }, []);
 
+  useEffect(() => {
+    Instabug.setNavigationListener(navigationRef);
+  }, [navigationRef]);
+
   return (
     <GestureHandlerRootView style={styles.root}>
       <NativeBaseProvider theme={nativeBaseTheme}>
         <QueryClientProvider client={queryClient}>
-          <NavigationContainer onStateChange={Instabug.onStateChange} theme={navigationTheme}>
+          <NavigationContainer theme={navigationTheme} ref={navigationRef}>
             <RootTabNavigator />
           </NavigationContainer>
         </QueryClientProvider>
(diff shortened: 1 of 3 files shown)


ahmedAlaaInstabug (approved):

ahmedAlaaInstabug:
good Job :clap:

test/modules/Instabug.spec.ts
```diff
@@ -236,6 +237,23 @@ describe('Instabug Module', () => {
...
+      goBack: jest.fn(),
+      dispatch: jest.fn(),
+      getRootState: jest.fn(),
+      canGoBack: jest.fn(),
+      addListener: jest.fn(),
+      removeListener: jest.fn(),
+    } as unknown as NavigationContainerRefWithCurrent<ReactNavigation.RootParamList>;
+
+    Instabug.setNavigationListener(mockNavigationContainerRef);
+    expect(mockNavigationContainerRef.addListener).toBeCalledTimes(1);
```
a7medev, line 254: I feel like this test is not really checking that the listener is attached correctly with the correct callback.

I think we should check that it attaches the listener to the `state` event, and that whenever the listener invokes the callback it reports the screen change correctly.

src/modules/Instabug.ts
```diff
@@ -533,6 +536,19 @@ export const onStateChange = (state?: NavigationStateV5) => {
...
+ * Sets a listener for screen change
+ *  @param navigationRef a refrence of a navigation container
+ *
+ */
+export const setNavigationListener = (
+  navigationRef: NavigationContainerRefWithCurrent<ReactNavigation.RootParamList>,
+) => {
+  navigationRef.addListener('state', () => {
+    onStateChange(navigationRef.getRootState());
+  });
```
a7medev, line 549: I think we should also return a function to unregister the listener. That way if for some reason the navigation ref changes the user can stop listening on the old one and listen on the new one instead.

So it could also be used like this with `useEffect`:
```js
useEffect(() => {
  return Instabug.setNavigationListener(navigationRef);
}, [navigationRef]);
```

examples/default/src/App.tsx
```diff
@@ -33,11 +34,15 @@ export const App: React.FC = () => {
...
+  useEffect(() => {
+    Instabug.setNavigationListener(navigationRef);
+  }, [navigationRef]);
+
   return (
     <GestureHandlerRootView style={styles.root}>
       <NativeBaseProvider theme={nativeBaseTheme}>
         <QueryClientProvider client={queryClient}>
-          <NavigationContainer onStateChange={Instabug.onStateChange} theme={navigationTheme}>
+          <NavigationContainer theme={navigationTheme} ref={navigationRef}>
```
a7medev, line 45: I think we can keep this one using `onStateChange` and add a new example app with Expo Router using the new API. What do you think?

a7medev (changes requested):
Too late to the review, but I'll leave the comments here and maybe we can address them in another PR.

examples/default/src/App.tsx
kholood-ea: I think adding expo example app is a good idea, but for now the navigation listener covers both cases.
we can change it back to onStateChange afterwards

a7medev: @kholood-ea I think having an Expo app in our examples would still be beneficial for testing other Expo-specific features. Besides it allows us to have both APIs (`Instabug.onStateChange` and `Instabug.setNavigationListener`) in the example apps. What do you think?

PR #1281, opened 2024-08-28
feat(android): add session sync callback

kholood-ea:
## Description of the change

Add a callback for session replay to determine weather to sync the session

## Type of change

## Related issues

[jiraID:MOB-15002](https://instabug.atlassian.net/browse/MOB-15002)
### Development

- [ ] Lint rules pass locally

### Code review

### DIFF
--- a/android/native.gradle
+++ b/android/native.gradle
@@ -1,5 +1,5 @@
 project.ext.instabug = [
-    version: '13.3.0'
+    version: '13.3.0.6212131-SNAPSHOT',
 ]
 
 dependencies {
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
@@ -116,7 +116,7 @@ public class RNInstabugSessionReplayModule extends EventEmitterModule {
 
     }
 
-    volatile boolean shouldSync = false;
+     boolean shouldSync = false;
      CountDownLatch latch;
     @ReactMethod
     public void setSyncCallback() {
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
@@ -172,6 +172,7 @@ public class RNInstabugSessionReplayModule extends EventEmitterModule {
                                 latch.await();
                             } catch (InterruptedException e) {
                                 e.printStackTrace();
+                                return true;
                             }
 
                             return shouldSync;
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
@@ -152,10 +152,10 @@ public class RNInstabugSessionReplayModule extends EventEmitterModule {
                             params.putString("device",sessionMetadata.getDevice());
                             params.putDouble("sessionDurationInSeconds",(double)sessionMetadata.getSessionDurationInSeconds());
                             params.putBoolean("hasLinkToAppReview",sessionMetadata.getLinkedToReview());
-                            params.putDouble("launchType", sessionMetadata.getLaunchType());
+                            params.putString("launchType",ArgsRegistry.launchTypeReversed.get(sessionMetadata.getLaunchType()) );
                             params.putDouble("launchDuration", sessionMetadata.getLaunchDuration());
                             params.putArray("networkLogs",getNetworkLogsArray(sessionMetadata.getNetworkLogs()));
-
+                            
 //                              TODO:Add rest of sessionMetadata
 //                            params.putDouble("bugsCount", ??);
 //                            params.putDouble("fatalCrashCount",??);
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
@@ -151,7 +151,7 @@ public class RNInstabugSessionReplayModule extends EventEmitterModule {
         return networkLogs;
     }
 
-    private boolean shouldSync = false;
+    private boolean shouldSync = true;
     private CountDownLatch latch;
     @ReactMethod
     public void setSyncCallback() {
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
@@ -115,9 +115,8 @@ public class RNInstabugSessionReplayModule extends EventEmitterModule {
 
 
     }
-
-     boolean shouldSync = false;
-     CountDownLatch latch;
+    private boolean shouldSync = false;
+    private CountDownLatch latch;
     @ReactMethod
     public void setSyncCallback() {
         MainThreadHandler.runOnMainThread(new Runnable() {
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
@@ -8,6 +8,7 @@ import com.facebook.react.bridge.Arguments;
 import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.WritableArray;
 import com.facebook.react.bridge.WritableMap;
 import com.instabug.library.OnSessionReplayLinkReady;
@@ -136,7 +137,7 @@ public class RNInstabugSessionReplayModule extends EventEmitterModule {
         return params;
     }
 
-    public WritableArray getNetworkLogsArray(List<SessionMetadata.NetworkLog> networkLogList ){
+    public ReadableArray getNetworkLogsArray(List<SessionMetadata.NetworkLog> networkLogList ){
         WritableArray networkLogs = Arguments.createArray();
 
         for (SessionMetadata.NetworkLog log : networkLogList) {
--- a/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
+++ b/android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
@@ -118,12 +118,28 @@ public class RNInstabugSessionReplayModule extends EventEmitterModule {
 
     }
 
+    public WritableMap getSessionMetadataMap(SessionMetadata sessionMetadata){
+        WritableMap params = Arguments.createMap();
+        params.putString("appVersion",sessionMetadata.getAppVersion());
+        params.putString("OS",sessionMetadata.getOs());
+        params.putString("device",sessionMetadata.getDevice());
+        params.putDouble("sessionDurationInSeconds",(double)sessionMetadata.getSessionDurationInSeconds());
+        params.putBoolean("hasLinkToAppReview",sessionMetadata.getLinkedToReview());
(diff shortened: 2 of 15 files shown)


android/src/test/java/com/instabug/reactlibrary/RNInstabugSessionReplayModuleTest.java
```diff
@@ -136,5 +135,51 @@ public void testSetUserStepsEnabled() {
...
     }
 
+    @Test
+
+    public void testSetSyncCallback() throws Exception  {
+        MockedStatic mockArgument = mockStatic(Arguments.class);
+        RNInstabugSessionReplayModule SRModule = spy(new RNInstabugSessionReplayModule(mock(ReactApplicationContext.class)));
+
+        CountDownLatch latch =new CountDownLatch(1);
+        SRModule.latch=latch;
```
a7medev, line 145: I don't think we should be setting the latch on `SRModule` directly in the test, `setSyncCallback` should be doing that internally.

Also, most likely we don't need to access the `CountDownLatch` in the test so let's mark it as `private`, if we must use it in the tests for some reason then let's mark it as `@VisibleForTesting`

android/src/test/java/com/instabug/reactlibrary/RNInstabugSessionReplayModuleTest.java
```diff
@@ -136,5 +135,51 @@ public void testSetUserStepsEnabled() {
...
+                });
+
+        Thread thread= new Thread (() ->{
+            try {
+                Thread.sleep(500);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+            SRModule.evaluateSync(true);
+        });
```
a7medev, line 165: I don't think we should use `Thread.sleep(500)` and instead restructure the test so that it executes without hacky delays.

The idea I have in mind is we'd post `onSessionReadyToSync` on a background thread (since that's how it's expected to be called by the Android SDK), call `setSyncCallback`, when the `IBG_SESSION_REPLAY_ON_SYNC_CALLBACK_INVOCATION` event is posted we call `evaluateSync` and after that we assert on the results.

What do you think?

android/src/test/java/com/instabug/reactlibrary/RNInstabugSessionReplayModuleTest.java
```diff
@@ -136,5 +135,51 @@ public void testSetUserStepsEnabled() {
...
+        SRModule.setSyncCallback();
+
+        WritableMap params = Arguments.createMap();
+        params.putString("appVersion","1.0");
+        params.putString("OS","android");
+        params.putString("device","device");
+        params.putDouble("sessionDurationInSeconds",20);
+
+       assertEquals(SRModule.shouldSync,true);
+       assertTrue("Latch should be zero after evaluateSync is called", SRModule.latch.getCount() == 0);
```
a7medev, line 177: I think these are internal implementation details we shouldn't be testing.

android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
```diff
@@ -97,6 +116,82 @@ public void onSessionReplayLinkReady(@Nullable String link) {
...
+                            params.putString("appVersion",sessionMetadata.getAppVersion());
+                            params.putString("OS",sessionMetadata.getOs());
+                            params.putString("device",sessionMetadata.getDevice());
+                            params.putDouble("sessionDurationInSeconds",(double)sessionMetadata.getSessionDurationInSeconds());
+                            params.putBoolean("hasLinkToAppReview",sessionMetadata.getLinkedToReview());
+                            params.putString("launchType",ArgsRegistry.launchTypeReversed.get(sessionMetadata.getLaunchType()) );
+                            params.putDouble("launchDuration", sessionMetadata.getLaunchDuration());
+                            params.putArray("networkLogs",getNetworkLogsArray(sessionMetadata.getNetworkLogs()));
+                            
+//                              TODO:Add rest of sessionMetadata
```
ahmedAlaaInstabug, line 159: What’s the status of the TODO tasks?

kholood-ea: it will be added in the next cycles as per the native team

android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
```diff
@@ -97,6 +116,82 @@ public void onSessionReplayLinkReady(@Nullable String link) {
...
+
+                            latch = new CountDownLatch(1);
+
+                            try {
+                                latch.await();
+                            } catch (InterruptedException e) {
+                                e.printStackTrace();
+                            }
+
+                            return shouldSync;
```
a7medev, line 174: Let's default to returning `true` if an exception is thrown in `latch.await()` or remove the `try-catch` and throw the exception to the native SDK to handle it if possible.

src/modules/SessionReplay.ts
```diff
@@ -1,5 +1,24 @@
...
+  device: string;
+  sessionDurationInSeconds: number;
+  networkLogs: Array<networkLog>;
+  launchType: LaunchType;
+  hasLinkToAppReview: boolean;
+  launchDuration: number;
+  bugsCount?: number;
+  fatalCrashCount?: number;
+  oomCrashCount?: number;
+}
```
a7medev, line 21: Let's rename these to `NetworkLog` and `SessionMetadata` to follow the naming convention for interfaces. Let's also move them to the `models` directory instead.

```suggestion
export interface NetworkLog {
  url: string;
  duration: number;
  statusCode: number;
}

export interface SessionMetadata {
  appVersion: string;
  OS: string;
  device: string;
  sessionDurationInSeconds: number;
  networkLogs: NetworkLog[];
  launchType: LaunchType;
  hasLinkToAppReview: boolean;
  launchDuration: number;
  bugsCount?: number;
  fatalCrashCount?: number;
  oomCrashCount?: number;
}
```

src/modules/SessionReplay.ts
```diff
@@ -75,3 +94,38 @@ export const setUserStepsEnabled = (isEnabled: boolean) => {
 export const getSessionReplayLink = async (): Promise<string> => {
   return NativeSessionReplay.getSessionReplayLink();
 };
+
+/**
+ * Set a callback for weather this session should sync
```
a7medev, line 99: Typo 🔎
```suggestion
 * Set a callback for whether this session should sync
```

src/modules/SessionReplay.ts
```diff
@@ -75,3 +94,38 @@ export const setUserStepsEnabled = (isEnabled: boolean) => {
...
+
+ * @example
+ * ```ts
+ * SessionReplay.setSyncCallback((payload)=>{
+ *    if(payload.device == "Xiaomi M2007J3SY" &&
+ *         payload.os == "OS Level 33" &&
+ *         payload.appVersion == "3.1.4 (4)" ||
+ *         payload.sessionDurationInSeconds > 20 )
+ *    {return true}
+ * });
```
a7medev, line 111:
```suggestion
 * SessionReplay.setSyncCallback((metadata) => {
 *    return metadata.device == "Xiaomi M2007J3SY" &&
 *         metadata.os == "OS Level 33" &&
 *         metadata.appVersion == "3.1.4 (4)" ||
 *         metadata.sessionDurationInSeconds > 20;
 * });
```

src/modules/SessionReplay.ts
```diff
@@ -75,3 +94,38 @@ export const setUserStepsEnabled = (isEnabled: boolean) => {
...
+    const result = handler(payload);
+    const shouldSync = Boolean(result);
+
+    if (typeof result !== 'boolean') {
+      console.warn(
+        `IBG-RN: The callback passed to SessionReplay.setSyncCallback was expected to return a boolean but returned "${result}". The value has been cast to boolean, proceeding with ${shouldSync}.`,
+      );
+    }
+
+    NativeSessionReplay.evaluateSync(Boolean(handler(payload)));
```
a7medev, line 127: We've already called the function and got the result above, no need to recall it here.

```suggestion
    NativeSessionReplay.evaluateSync(shouldSync);
```

src/native/NativeSessionReplay.ts
```diff
@@ -1,4 +1,4 @@
-import type { NativeModule } from 'react-native';
+import { NativeEventEmitter, type NativeModule } from 'react-native';
```
a7medev, line 1: Let's split the regular import and the type-only import to avoid issues with older versions of TypeScript that don't support having both type-only and regular imports in the same import statement.

```suggestion
import { NativeEventEmitter } from 'react-native';
import type { NativeModule } from 'react-native';
```

test/modules/SessionReplay.spec.ts
```diff
@@ -36,4 +36,15 @@ describe('Session Replay Module', () => {
...
+
+  it('should call the native method setSyncCallback', () => {
+    const callback = jest.fn().mockReturnValue(true);
+
+    SessionReplay.setSyncCallback(callback);
+    emitter.emit(NativeEvents.SESSION_REPLAY_ON_SYNC_CALLBACK_INVOCATION);
+
+    expect(NativeSessionReplay.setSyncCallback).toBeCalledTimes(1);
+    expect(emitter.listenerCount(NativeEvents.SESSION_REPLAY_ON_SYNC_CALLBACK_INVOCATION)).toBe(1);
+    expect(NativeSessionReplay.evaluateSync).toBeCalledTimes(1);
```
a7medev, line 48: Let's assert that `evaluateSync` gets called with the correct value as well.

```suggestion
    const shouldSync = true;
    const callback = jest.fn().mockReturnValue(shouldSync);

    SessionReplay.setSyncCallback(callback);
    emitter.emit(NativeEvents.SESSION_REPLAY_ON_SYNC_CALLBACK_INVOCATION);

    expect(NativeSessionReplay.setSyncCallback).toBeCalledTimes(1);
    expect(emitter.listenerCount(NativeEvents.SESSION_REPLAY_ON_SYNC_CALLBACK_INVOCATION)).toBe(1);
    expect(NativeSessionReplay.evaluateSync).toBeCalledTimes(1);
    expect(NativeSessionReplay.evaluateSync).toBeCalledWith(shouldSync);
```

a7medev (changes requested):
Good job @kholood-ea 👏🏼
Please check the comments added, I'd love to hear your thoughts 🙏🏼

android/native.gradle
```diff
@@ -1,5 +1,5 @@
 project.ext.instabug = [
-    version: '13.3.0'
+    version: '13.3.0.6212131-SNAPSHOT',
```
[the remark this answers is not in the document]
kholood-ea: I believe it hasn't been just yet.

android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
a7medev: @kholood-ea We need to resolve this one before we merge the PR to avoid unintentional return values from the sync callback.

android/src/main/java/com/instabug/reactlibrary/RNInstabugSessionReplayModule.java
```diff
@@ -97,6 +117,84 @@ public void onSessionReplayLinkReady(@Nullable String link) {
             }
         });
 
+    }
+
+    public WritableMap getSessionMetadataMap(SessionMetadata sessionMetadata){
```
a7medev, line 122:
```suggestion
    public ReadableMap getSessionMetadataMap(SessionMetadata sessionMetadata){
```

src/modules/SessionReplay.ts
```diff
@@ -1,5 +1,24 @@
...
+  device: string;
+  sessionDurationInSeconds: number;
+  networkLogs: NetworkLog[];
+  launchType: LaunchType;
+  hasLinkToAppReview: boolean;
+  launchDuration: number;
+  bugsCount?: number;
+  fatalCrashCount?: number;
+  oomCrashCount?: number;
+}
```
a7medev, line 21: Let's move them to the `models` directory please.

a7medev (changes requested):

AyaMahmoud148 (approved):

PR #1305, opened 2024-10-31
ci: validate release post check job

AyaMahmoud148:
## Description of the change

> 
1.  Packed the sdk to be use as npm package instead of linking with source code 
2. add the package to the hybrid app
3. Build the hybrid app with the SDK package to make sure that everything is oki

## Type of change

- [X] New feature (non-breaking change that adds functionality)

## Related issues

> https://instabug.atlassian.net/browse/MOB-16270

### Development

- [X] Lint rules pass locally
- [X] The code changed/added as part of this pull request has been covered with tests

### Code review

- [X] This pull request has a descriptive title and information useful to a reviewer
- [X] Issue from task tracker has a link to this pull request


.circleci/config.yml
```diff
@@ -154,6 +154,22 @@ commands:
       - run:
           name: Post comment on GitHub
           command: ./scripts/notify-github.sh "<< parameters.data >>"
+  pack_and_build_sdk:
```
a7medev, line 157: I feel like the order of `build` then `pack` makes more sense for the name, what do you think?
```suggestion
  build_and_pack_sdk:
```

.circleci/config.yml
```diff
@@ -154,6 +154,22 @@ commands:
...
+    steps:
+      - install_node_modules
+      - run:
+          command: yarn build
+          working_directory: ~/project
+      - run:
+          command: yarn pack --filename InstabugSDK.tgz
+          working_directory: ~/project
+      - run:
+          command: yarn add file:../../InstabugSDK.tgz
```
a7medev, line 171: Nice! I didn't know we could install from a `tgz` file

.circleci/config.yml
```diff
@@ -154,6 +154,22 @@ commands:
...
+          working_directory: ~/project/examples/hybrid
```
a7medev, line 172: Do we need to lock this to the hybrid example app? We might want to integrate the same command in other example apps in the future, what do you think?
We can probably use the `working_directory` passed for this and default the `working_directory` parameter to the `hybrid` example app.

```suggestion
          working_directory: <<parameters.working_directory>>
```

.circleci/config.yml
```diff
@@ -487,6 +503,38 @@ jobs:
...
+      name: android/android-machine
+      tag: '2024.01.1'
+    steps:
+      - advanced-checkout/shallow-checkout
+      - pack_and_build_sdk
+      - run:
+          name: Build Android
+          command: |
+            cd examples/hybrid/android
+            ./gradlew assembleDebug
```
a7medev, line 518: I believe this job as well doesn't have to be restricted to the hybrid example app. We can also build the app in release mode instead. I also feel like `build_check_android` would be a more suitable name, what do you think?

```suggestion
  build_check_android:
    working_directory: ~/project
    executor:
      name: android/android-machine
      tag: '2024.01.1'
    parameters:
      app_dir:
        type: string
        default: examples/hybrid
    steps:
      - advanced-checkout/shallow-checkout
      - pack_and_build_sdk
      - run:
          name: Build Android
          command: ./gradlew assembleRelease
          working_directory: <<parameters.app_dir>>/android
```

.circleci/config.yml
```diff
@@ -487,6 +503,38 @@ jobs:
...
+
+  build_ios_release:
```
a7medev, line 520: Same comment as for the Android one.

examples/hybrid/package.json
a7medev, line 1: I feel like we can default to using the symlinked `instabug-reactnative` package in the `hybrid` app for ease of development. The CI jobs can always install the version from a pack if they need to but in development we'd usually need to use the local package and have the modifications show up instantly, what do you think?

a7medev (changes requested):
Good job @AyaMahmoud148! I've got some comments, please check them out, I'd love to hear your thoughts.

examples/hybrid/package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@react-native-community/cli": "^14.1.1",
     "@react-native-community/cli-platform-android": "^14.1.1",
-    "instabug-reactnative": "^13.4.2",
+    "instabug-reactnative": "file:../../InstabugSDK.tgz",
```
a7medev, line 15: I think it needs to be `link:../..` like we have in the `default` example app. Having it `file:../../InstabugSDK.tgz` would require us to build and pack the SDK with every change to reflect in the hybrid app.

.circleci/config.yml
```diff
@@ -487,6 +503,40 @@ jobs:
...
+      tag: '2024.01.1'
+    steps:
+      - advanced-checkout/shallow-checkout
+      - build_and_pack_sdk:
+          working_directory: ~/project/examples/hybrid
+      - run:
+          name: Build Android
+          command: |
+            cd examples/hybrid/android
+            ./gradlew assembleDebug
```
a7medev, line 519: Let's also make this job generic for different example apps and build the app in release mode as mentioned in the previous comment. Let's do the same for `build_check_ios` as well.

package.json
```diff
@@ -36,7 +36,8 @@
     "example": "yarn --cwd examples/default",
     "hybrid": "yarn --cwd examples/hybrid",
     "pods": "cd examples/default &&  pod-install --quiet && cd ../hybrid && pod-install --quiet ",
-    "bootstrap": "yarn example && yarn hybrid && yarn && yarn pods"
+    "packSDK": "yarn build && yarn pack --filename InstabugSDK.tgz",
+    "bootstrap": "yarn packSDK && yarn example && yarn hybrid && yarn && yarn pods"
```
a7medev, line 40: We probably don't need `yarn packSDK` in the `bootstrap` scripts, it's only needed when we want to build the app for a build check like we do on CI.

package.json
```diff
@@ -36,7 +36,8 @@
     "example": "yarn --cwd examples/default",
     "hybrid": "yarn --cwd examples/hybrid",
     "pods": "cd examples/default &&  pod-install --quiet && cd ../hybrid && pod-install --quiet ",
-    "bootstrap": "yarn example && yarn hybrid && yarn && yarn pods"
+    "packSDK": "yarn build && yarn pack --filename InstabugSDK.tgz",
```
a7medev, line 39: We can probably rename this to something like `pack:sdk` rather than `packSDK` to match the naming convention for the rest of the scripts, what do you think?

```suggestion
    "pack:sdk": "yarn build && yarn pack --filename InstabugSDK.tgz",
```

a7medev (changes requested):

.circleci/config.yml
```diff
@@ -487,6 +503,51 @@ jobs:
           replace: $SNAPSHOT_VERSION
       - notify_github:
           data: "$(jq -Rcs '{ body: . }' scripts/snapshot-comment.md)"
+  build_check_android:
+    parameters:
+      working_directory:
+        type: string
+      android_path:
+        type: string
```
a7medev, line 511: I feel like `android_path` is always `<<parameters.working_directory>>/android`, do we need to pass it as a parameter as well?

.circleci/config.yml
```diff
@@ -487,6 +503,51 @@ jobs:
...
+      tag: '2024.01.1'
+    steps:
+      - advanced-checkout/shallow-checkout
+      - build_and_pack_sdk:
+          working_directory: << parameters.working_directory >>
+      - run:
+          name: Build Android
+          command: |
+            cd << parameters.android_path >>
+            ./gradlew assembleRelease
```
a7medev, line 523: Besides getting the Android path from the working directory parameter, I believe using CircleCI config's `working_directory` instead of `cd` in this case would be more readable. So in total I feel like it should look like this:
```suggestion
          command: ./gradlew assembleRelease
          working_directory: <<parameters.working_directory>>/android
```

.circleci/config.yml
```diff
@@ -487,6 +503,51 @@ jobs:
...
+
+  build_check_ios:
+    parameters:
+      working_directory:
+        type: string
+      ios_path:
+        type: string
```
a7medev, line 530: Same comment as `android_path`

.circleci/config.yml
```diff
@@ -487,6 +503,51 @@ jobs:
...
+          name: Build iOS
+          command: |
+            cd << parameters.ios_path >>
+            yarn react-native bundle --platform ios --dev false --entry-file index.js --bundle-output ios/main.jsbundle --assets-dest ios
+            rm -rf ~/Library/Developer/Xcode/DerivedData/*
+            xcodebuild -workspace HybridSampleApp.xcworkspace \
+            -scheme HybridSampleApp \
+            -configuration Release \
+            -sdk iphonesimulator \
+            -destination "platform=iOS Simulator,OS=15.5,name=iPhone 11 Pro Max"
```
a7medev, line 550: Same comment as the Android one.

.circleci/config.yml
```diff
@@ -487,6 +503,51 @@ jobs:
...
+      - build_and_pack_sdk:
+          working_directory: << parameters.working_directory >>
+      - install_pods:
+          working_directory: << parameters.ios_path >>
+      - run:
+          name: Build iOS
+          command: |
+            cd << parameters.ios_path >>
+            yarn react-native bundle --platform ios --dev false --entry-file index.js --bundle-output ios/main.jsbundle --assets-dest ios
+            rm -rf ~/Library/Developer/Xcode/DerivedData/*
```
a7medev, line 545: Why do we need to remove the `DerivedData` on CI? 👀 Is that due to some flakiness?

a7medev (changes requested):

.circleci/config.yml
AyaMahmoud148: yes the ci was randomly keeping failing with the drived data

a7medev (approved):
Great job @AyaMahmoud148! 👏🏼❤️

PR #1306, opened 2024-11-03
Feat/network spans filtering

AndrewAminInstabug:
## Description of the change

> Add obfuscation and request filtering logic for native interception 
## Type of change

## Related issues

Jira ID: APM-20927

### Development

- [x] Lint rules pass locally

### Code review


ahmedAlaaInstabug (approved):

ios/RNInstabug/InstabugNetworkLoggerBridge.m
```diff
@@ -44,6 +73,145 @@ -(void)stopObserving {
...
+        ((IBGURLRequestAsyncObfuscationCompletedHandler)self.requestObfuscationCompletionDictionary[callbackID])(request);
+    } else {
+        NSLog(@"Not Available Completion");
+    }
+
+}
+
+RCT_EXPORT_METHOD(setNetworkLoggingRequestFilterPredicateIOS: (NSString * _Nonnull) callbackID : (BOOL)value ){
+
+    if ([callbackID isKindOfClass:[NSString class]] && self.requestFilteringCompletionDictionary[callbackID] != nil) {
```
a7medev, line 141: I assume `callbackID` will always be a non-null `NSString *` given the method signature, why do we need the check for its class? 🤔
```suggestion
    if (self.requestFilteringCompletionDictionary[callbackID] != nil) {
```

ios/RNInstabug/InstabugNetworkLoggerBridge.m
```diff
@@ -44,6 +73,145 @@ -(void)stopObserving {
...
+        // ⬇️ YES == Request will be saved, NO == will be ignored
+        ((IBGURLRequestResponseAsyncFilteringCompletedHandler)self.requestFilteringCompletionDictionary[callbackID])(value);
```
a7medev, line 143: Why do we need to explicitly cast the value from the dictionary when it already has this type specified?

src/modules/Instabug.ts
```diff
@@ -99,29 +101,24 @@ function _logFlags() {
...
-  addAppStateListener(async (nextAppState) => handleAppStateChange(nextAppState, config));
+    // Add android feature flags listener for android
+    addOnFeatureUpdatedListener(config);
+  } else {
+    isNativeInterceptionFeatureEnabled = await NativeNetworkLogger.isNativeInterceptionEnabled();
 
-  // Perform platform-specific checks and update interception mode
-  handleNetworkInterceptionMode(config);
+    // Add app state listener to handle background/foreground transitions
+    addAppStateListener(async (nextAppState) => handleAppStateChange(nextAppState, config));
```
a7medev, line 111: TBH I don't think listening manually on app state change (background/foreground) is a good idea for making sure that the feature flags are in sync for the same reason as why the approach didn't work on Android. I think it'd be better to follow the same approach with a listener like on Android to avoid the feature flags getting out of sync between native and JS.

src/modules/NetworkLogger.ts
```diff
@@ -131,3 +161,87 @@ export const apolloLinkRequestHandler: RequestHandler = (operation, forward) =>
...
+        contentType: '',
+        duration: 0,
+        requestBodySize: 0,
+        responseBodySize: 0,
+        errorDomain: '',
+        errorCode: 0,
+        startTime: 0,
+        serverErrorMessage: '',
+        requestContentType: '',
+      };
```
a7medev, line 236: **comment**: I feel like with the new approach the obfuscation/filtering handlers don't get a `NetworkData` object anymore making this type misleading.

a7medev (changes requested):
Great effort @AndrewAminInstabug 👏🏼
I've left some comments that I'd love to hear your thoughts on.

src/modules/NetworkLogger.ts
AndrewAminInstabug: We used the same type currently implemented in production to avoid breaking the existing API. I recall we previously discussed this and ultimately decided to proceed with the existing model.

src/modules/Instabug.ts
AndrewAminInstabug: @a7medev, AWU, however this implementation is done due to and iOS limitation.

android/src/main/java/com/instabug/reactlibrary/RNInstabugNetworkLoggerModule.java
```diff
@@ -81,4 +104,70 @@ public void run() {
...
+            }
+        });
+    }
+
+    @ReactMethod
+    public void resetNetworkLogsListener() {
+        MainThreadHandler.runOnMainThread(new Runnable() {
+            @Override
+            public void run() {
+                InternalAPM._registerNetworkLogSanitizer(null);
```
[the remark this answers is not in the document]
AndrewAminInstabug: per the android team, setting the sanitizer to `null` would be enough 👍

ios/RNInstabug/InstabugNetworkLoggerBridge.m
AndrewAminInstabug: This code snippet was created by **Eyad** from the iOS team, and this is how it worked for them.

ios/RNInstabug/InstabugNetworkLoggerBridge.m
AndrewAminInstabug: Same as the above snippet

ios/RNInstabug/InstabugNetworkLoggerBridge.m
```diff
@@ -44,6 +73,145 @@ -(void)stopObserving {
...
+
+         case NetworkListenerTypeBoth:
+             NSLog(@"Andrew: Both filtering and obfuscation enabled");
+             [self setupRequestFilteringAndObfuscationHandler];
+             break;
+
+         default:
+             NSLog(@"Andrew: Unknown NetworkListenerType");
+             break;
+     }
```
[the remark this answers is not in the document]
AndrewAminInstabug: The logic is divided into separate methods and consolidated within this method. The method's responsibility is to determine which specific method to invoke.

android/src/main/java/com/instabug/reactlibrary/RNInstabugNetworkLoggerModule.java
```diff
@@ -81,4 +104,70 @@ public void run() {
...
+        try {
+            newJSONObject = new JSONObject(jsonString);
+            url = newJSONObject.optString("url");
+        } catch (JSONException e) {
+            throw new RuntimeException(e);
+        }
+        final String ID = newJSONObject.optString("id");
+
+        if (!url.isEmpty()) {
+            modifiedSnapshot = new NetworkLogSnapshot(url, null, null, null, null, newJSONObject.optInt("responseCode"));
```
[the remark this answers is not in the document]
AndrewAminInstabug: In APM, the native side only interests in url and timings, any other data is ignored.

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
```diff
@@ -1167,4 +1170,20 @@ public Map<String, Object> getConstants() {
...
         return constants;
     }
+
+    @ReactMethod
+    public void setOnFeaturesUpdatedListener() {
+        InternalCore.INSTANCE._setOnFeaturesUpdatedListener(new OnFeaturesUpdatedListener() {
+            @Override
+            public void invoke() {
+                final boolean cpNativeInterceptionEnabled = InternalAPM._isFeatureEnabledCP(CP_NATIVE_INTERCEPTION_ENABLED, "");
+                final boolean hasAPMPlugin = InternalAPM._isFeatureEnabledCP(APM_NETWORK_PLUGIN_INSTALLED, "");
```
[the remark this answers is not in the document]
AndrewAminInstabug: This section is overseen by the Android team. However I believe you're right.

android/src/main/java/com/instabug/reactlibrary/RNInstabugNetworkLoggerModule.java
```diff
@@ -39,7 +52,17 @@ public void removeListeners(Integer count) {
...
 
     private boolean getFlagValue(String key) {
-        return InternalAPM._isFeatureEnabledCP( key , "");
+        return InternalAPM._isFeatureEnabledCP(key, "");
+    }
+
+    private WritableMap convertFromMapToWriteableMap(Map map) {
+        WritableMap writableMap = new WritableNativeMap();
+        for (int i = 0; i < map.size(); i++) {
+            Object key = map.keySet().toArray()[i];
```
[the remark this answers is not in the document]
AndrewAminInstabug: Do you mean as follow ??

```
    private WritableMap convertFromMapToWritableMap(Map<String, Object> map) {
        WritableMap writableMap = new WritableNativeMap();
        for (String key : map.keySet()) {
            writableMap.putString(key, (String) map.get(key));
        }
        return writableMap;
    }
```

ios/RNInstabug/InstabugNetworkLoggerBridge.m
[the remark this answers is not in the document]
a7medev: I mean why not just call these methods directly instead of calling this method with the enum value for it 🤔

android/src/main/java/com/instabug/reactlibrary/RNInstabugNetworkLoggerModule.java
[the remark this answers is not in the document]
a7medev: Alright. Just to confirm, in core will all the data still be available?

android/src/main/java/com/instabug/reactlibrary/RNInstabugNetworkLoggerModule.java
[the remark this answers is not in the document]
a7medev: @AndrewAminInstabug Yes I think that will work fine, no?

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
```diff
@@ -1167,4 +1170,20 @@ public Map<String, Object> getConstants() {
 
         return constants;
     }
+
+    @ReactMethod
+    public void setOnFeaturesUpdatedListener() {
+        InternalCore.INSTANCE._setOnFeaturesUpdatedListener(new OnFeaturesUpdatedListener() {
```
ahmedAlaaInstabug, line 1176: what's the different between _setOnFeaturesUpdatedListener and _setFeaturesStateListener ?

src/modules/Instabug.ts
```diff
@@ -247,25 +238,23 @@ function handleAndroidNativeInterception() {
...
-    console.log(
-      `Andrew: handleIOSNativeInterception(${
-        shouldEnableNativeInterception &&
-        config.networkInterceptionMode === NetworkInterceptionMode.native
-      })`,
-    );
-
-    if (
+  console.log(
+    `Andrew: handleIOSNativeInterception(${
```
ahmedAlaaInstabug, line 245: I think this log should be removed.

src/modules/Instabug.ts
```diff
@@ -322,6 +311,47 @@ const initializeNativeInstabug = (config: InstabugConfig) => {
...
+  }
+}
+
+/**
+ * Add Android Listener for native feature flags changes.
+ */
+function addOnFeatureUpdatedListener(config: InstabugConfig) {
+  emitter.addListener(NativeEvents.IBG_ON_FEATURES_UPDATED_CALLBACK, (flags) => {
+    const { cpNativeInterceptionEnabled, hasAPMPlugin } = flags;
+    console.log(`Andrew: addOnFeatureUpdatedListener ->
```
ahmedAlaaInstabug, line 342: i think this console log should be removed

ahmedAlaaInstabug (commented):
Impressive work! 👏 👏
I don't have any comments other than @a7medev comments. You can check the console log statements, as I believe they were only useful for debugging purposes.

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
AndrewAminInstabug: Honestly, I'm not familiar with `_setFeaturesStateListener`; Android provides `_setOnFeaturesUpdatedListener` specifically for network flag updates.

src/modules/Instabug.ts
AndrewAminInstabug: There is a **TODO** for removing logs. 👍

ios/RNInstabug/InstabugNetworkLoggerBridge.m
a7medev: I confirmed with Eyad and he agreed that it's not needed. I think we can remove it.

ios/RNInstabug/InstabugNetworkLoggerBridge.m
a7medev: I confirmed with Eyad and he agreed that it's not needed. I think we can remove it.

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
AndrewAminInstabug: Andrew Safwat from the Android team confirmed that the `_setOnFeaturesUpdatedListener` will be the new listener the team will maintain in the future. This change is because the `_setFeaturesStateListener` lacks extensibility; for instance, APM cannot utilize it.

android/src/main/java/com/instabug/reactlibrary/RNInstabugReactnativeModule.java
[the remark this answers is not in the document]
AndrewAminInstabug: Andrew Safwat from the Android team confirmed that including the caller name here results in unhelpful error messages for debugging. He recommended sending it as empty strings instead.

a7medev (approved):
Great job @AndrewAminInstabug 👏🏼👏🏼

PR #1307, opened 2024-11-06
feat: exclude DEV server from network logs

ahmedAlaaInstabug:
## Description of the change

1) feat: exclude DEV server from network logs

## Type of change

## Related issues

JIRA ID :  MOB-16570

### Development

- [ ] Lint rules pass locally

### Code review


a7medev (approved):
LGTM 👌🏼

src/modules/NetworkLogger.ts
```diff
@@ -2,14 +2,28 @@ import type { RequestHandler } from '@apollo/client';
...
 let _networkDataObfuscationHandler: NetworkDataObfuscationHandler | null | undefined;
 let _requestFilterExpression = 'false';
 
+const filterURL = (url: string): string => {
+  const [protocol, rest] = url.split('://');
+  return (rest || protocol).split('/')[0];
+};
+
+const getDevServerURL = (): string | null => {
+  const { scriptURL } = NativeModules.SourceCode;
```
a7medev, line 20: Is there a higher level API that we can use instead of using `NativeModules.SourceCode`? If not, I believe we need to check first if `SourceCode` is not `null` or `undefined` before trying to access `scriptURL` on it.

src/modules/NetworkLogger.ts
```diff
@@ -2,14 +2,28 @@ import type { RequestHandler } from '@apollo/client';
...
+  if (scriptURL) {
+    return filterURL(new URL(scriptURL).toString());
```
a7medev, line 22: Why do we need to use `new URL(..).toString()`? Does it do any kind of normalization for the `scriptURL`?

a7medev (changes requested):

src/modules/NetworkLogger.ts
ahmedAlaaInstabug: fixed

src/modules/NetworkLogger.ts
ahmedAlaaInstabug: I didn't find any other API, but I added a condition to check first.

src/modules/NetworkLogger.ts
```diff
@@ -2,14 +2,30 @@ import type { RequestHandler } from '@apollo/client';
...
 let _networkDataObfuscationHandler: NetworkDataObfuscationHandler | null | undefined;
 let _requestFilterExpression = 'false';
 
+const filterURL = (url: string): string => {
+  const [protocol, rest] = url.split('://');
+  return (rest || protocol).split('/')[0];
+};
+
+const getDevServerURL = (): string | null => {
+  if (NativeModules.SourceCode) {
```
kholood-ea, line 20: this probably wont work with the new architecture how about adding a fallback something like this :

`import Config from 'react-native-config';`
`const configURL = Config.DEV_SERVER_URL;`  
`if (configURL) {return configURL}`

what do you think ?

kholood-ea (approved):

src/modules/NetworkLogger.ts
a7medev: @kholood-ea This is a very good point that we might need to validate.
Though I don't think we can use `react-native-config` in our SDK since it's not a dependency of our SDK (and we don't intend to add it), but we can probably check its implementation in `react-native-config` and use a similar approach if that's doable.
@ahmedAlaaInstabug What do you think?

ahmedAlaaInstabug: This code doesn’t work in the new architecture setup :( It only works if the user specifies the dev server in the **.env** file.

a7medev: @ahmedAlaaInstabug I believe we need to find an approach that works with the new architecture as well, what do you think?

ahmedAlaaInstabug: I updated the logic to check if the URL port matches the Metro port, and if so, it will be excluded. Additionally, I added a new API to set the Metro port if it’s changed from the default 8081.

AndrewAminInstabug (dismissed):

src/modules/Instabug.ts
```diff
@@ -633,6 +633,13 @@ export const willRedirectToStore = () => {
   NativeInstabug.willRedirectToStore();
 };
 
+/**
+ * This API has be called when changing the default Metro server port (8081) to exclude the DEV URL from network logging.
+ */
+export const setMetroDevServerPort = (port: number) => {
+  InstabugConstants.METRO_SERVER_URL = port.toString();
```
a7medev, line 640: I'd say using let's move the Metro server port constant from `InstabugConstants` to a variable since it's not really "constant", what do you think? We can keep the default port (`8081`) in `InstabugConstants` though but not mutate it.

a7medev (changes requested):

src/utils/InstabugConstants.ts
```diff
@@ -11,6 +11,7 @@ const InstabugConstants = {
     'IBG-RN: Expected key and value passed to setUserAttribute to be of type string',
   REMOVE_USER_ATTRIBUTES_ERROR_TYPE_MESSAGE:
     'IBG-RN: Expected key and value passed to removeUserAttribute to be of type string',
+  DEFAULT_METRO_SERVER_URL: '8081',
```
kholood-ea, line 14: I believe this
```suggestion
  DEFAULT_METRO_PORT: '8081',
``` 
would make more sense here what do you think ?

a7medev (approved):

kholood-ea (dismissed):

AyaMahmoud148 (dismissed):

AyaMahmoud148 (approved):

AndrewAminInstabug (approved):

